PR #1, opened 2020-04-01
Varying beta

richfitz:
This PR adds a time-varying beta, following @MJomaba's implementation. There is no test that it behaves as expected, nor that issues around when the time an intervention takes place is correct (is it at the start or the end of the day? check the odin docs and compare with the model output).

@johnlees - the files R/odin.R, src/odin.c and inst/odin/basic.json are automatically generated from inst/odin/basic.R


R/parameters.R
```diff
@@ -180,9 +180,8 @@ generate_parameters <- function(
                          D0 = R0,
                          trans_increase = trans_increase_array,
                          trans_profile = trans_profile_array,
-                         beta_list = beta,
-                         beta = beta[1], # When odin code is updated, set this list
-                         beta_dates = beta_dates,
+                         beta_y = beta,
```
johnlees, line 183: Is it worth adding in a similar set to `NULL` in `test-parameters.R` to ignore the betas for the time being, and still get the tests passing. Looks like all the others are fine

R/parameters.R
```diff
@@ -180,9 +180,8 @@ generate_parameters <- function(
...
+                         beta_t = beta_dates,
```
johnlees, line 184: Just double-checking the definition at https://github.com/mrc-ide/sircovid/blob/master/R/parameters.R#L100 is compatible with the rest of the odin code. Unix time stamps differences (days) from the first date entered here

R/parameters.R
richfitz: Yup, added that now

R/parameters.R
richfitz: I think that this is correct but @MJomaba needs to confirm. I feel we're very much at risk of off-by-one errors here, but the conversion looked correct to within one day to me

R/parameters.R
MJomaba: Shouldn't it be D0=D0 rather R0?

R/parameters.R
MJomaba: I think it is fine

R/parameters.R
richfitz: Nice spot Marc, though of course they're the same here at the moment so this had no effect.  Fixed now

PR #2, opened 2020-04-01
Package infrastructure

richfitz:
This PR adds a few nice-to-haves:

* code coverage, though we ignore the generated odin.c file
* a pkgdown website (currently manually updated but available here: https://mrc-ide.github.io/sircovid/)
* fix a few relatively harmless notes from R CMD check

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -17,3 +17,5 @@
 ^README\.md\.in$
 ^pkgapi_.*\.tar\.gz$
 ^\.travis\.yml$
+^codecov\.yml$
+^\.covrignore$
--- a/.covrignore
+++ b/.covrignore
@@ -1,0 +1,1 @@
+src/odin.c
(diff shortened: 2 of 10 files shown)


.Rbuildignore
```diff
@@ -17,3 +17,5 @@
 ^README\.md\.in$
 ^pkgapi_.*\.tar\.gz$
 ^\.travis\.yml$
+^codecov\.yml$
+^\.covrignore$
```
johnlees, line 21: Useful to add `*.Rproj`, or is there a good reason for not doing that? I have `sircovid.Rproj` which I'd prefer to be ignored

richfitz: that should be added to the Rbuildignore, yeah.  Elsewhere I have

```
^.*\.Rproj$
^\.Rproj\.user$
```

for Rstudio files.  I'll add them now, but we should commit the project file too - could you push it to this branch please?

johnlees: Will do, then I will merge this

PR #8, opened 2020-04-02
Grid search

### LINKED ISSUE #3: Simple grid search
Port across the code from ncov-outputs - this is the MVP for the real-time reporting

johnlees:
This adds a function to run a simple grid search across beta and start date ranges, which run in parallel. Function make simple plots of log-l and probability, and returns grid and probability values.

Tests check structure of results is ok, and does a simple test of beta and date ranges to check reasonable values are preferred.

I haven't put in a test of the plots, if that's needed I would probably need some advice

This closes #3


johnlees:
Drop in coverage seems mostly because codecov doesn't understand the `furrr` execution

R/grid_search.R
```diff
@@ -0,0 +1,124 @@
...
+    if (length(model_params$beta_y)) {
+      stop("Time-varying beta not yet supported by grid search")
+    }
+  }
+  
+  #
+  # Multi-core futures with furrr (parallel purrr)
+  #
+  ## Particle filter outputs
+  plan(multiprocess)
```
richfitz, line 81: I think that this should come out, and be part of user control. That deals with the coverage issue because the user then sets it to be in parallel or not

R/grid_search.R
```diff
@@ -0,0 +1,124 @@
...
+  pf_run_outputs <- furrr::future_pmap(
+    param_grid,
+    function(beta, start_date) {
```
richfitz, line 84: This looks generalisable to me, but I can look at that for the next stage (given other things to search over)

R/grid_search.R
```diff
@@ -0,0 +1,124 @@
...
+    }
+  )
+  ## Extract Log-likelihoods
+  log_ll <- furrr::future_map(pf_run_outputs, ~ .$log_likelihood) 
+  
+  ## Construct a matrix with start_date as columns, and beta as rows
+  ## order of return is set by order passed to expand.grid, above
+  ## Returned column-major (down columns of varying beta) - set byrow = FALSE
+  mat_log_ll <- matrix(
+    as.numeric(log_ll),
```
richfitz, line 104: As we getting a list out? If so you might prefer `future_map_dbl`

R/grid_search.R
```diff
@@ -0,0 +1,124 @@
...
+##' @param pars_obs list of parameters to use in comparison
+##'   with \code{compare_icu}.
+##'   
+##' @return List of beta and start date grid values, and
+##'   normalised probabilities at each point
+##' 
+##' @export
+##' @import furrr
+##' @import future
+scan_beta_date <- function(
```
richfitz, line 33: OK, I think we can generalise this here: how about:

```
scan_likelihood <- function(pars, data, model_params, pars_obs) {
  param_grid <- do.call("expand.grid", pars)
  ...
}
```

Then we need to generalise the approach below.  Or we can proceed with this and I can pop that generalisation together later and get you to review?

R/grid_search.R
```diff
@@ -0,0 +1,124 @@
...
+  # Set up parameter space to scan
+  #
+  beta_1D <- seq(min_beta, max_beta, beta_step)
+  date_list <- seq(as.Date(first_start_date), as.Date(last_start_date), day_step)
+  param_grid <- expand.grid(beta = beta_1D, start_date = date_list)
+  
+  #
+  # Set up calls to simulator runs
+  #
+  if (is.null(model_params)) {
```
richfitz, line 57: Can you give pars_obs the same treatment to simplify the signature?

R/grid_search.R
```diff
@@ -0,0 +1,124 @@
...
+    as.numeric(log_ll), 
+    nrow = length(beta_1D),
+    ncol = length(date_list),
+    byrow = FALSE)
+
+  # Exponentiate elements and normalise to 1 to get probabilities
+  prob_matrix <- exp(mat_log_ll)
+  renorm_mat_LL <- prob_matrix/sum(prob_matrix)
+  
+  # Display likelihood/probability surfaces across grid
```
richfitz, line 113: "A function should either compute something or do something" (doing something being a side effect.

Can you drop this for now?  The R-ish way to do this might be closer to:

```
class(results) <- "sircovid_scan"
results
```

then a new plot method:

```
##' @export
plot.sircovid_scan <- function(x, ..., what = "likelihood") {
  if (what == "likelihood") {
    graphics::image(x=x$x, y=x$y, z=x$mat_log_ll,
          xlab="beta", ylab="Start date", main = "Log-likelihood")
   } # ...etc, with some liberties taken for generalisations....
}
```

R/grid_search.R
johnlees: It's not immediately obvious to me how to generalise over possible changes in all of the input parameters. Named members of `model_params` would be ok, but what about members of `pars_obs`. Do we need to add a few conditionals in here to check what is changing, or is there a cleverer way?

R/grid_search.R
johnlees: Yep, this is better

R/grid_search.R
johnlees: Removed

johnlees:
I've done the small changes, but wasn't totally sure which parameters might need to be scanned. Happy to either implement your suggestion, or hand over to you for that part. The output and plotting is at least generalised with `x` and `y`

R/grid_search.R
richfitz: I can throw something together (#9) and you can look it over - I think this will turn up again for the other methods too...

PR #10, opened 2020-04-02
Severity file parsing improvement

### LINKED ISSUE #4: Better processing of the severity file
This is very fragile and it will be hard to incorporate a new one.  Check things like the variable names, and work out what else has to match up.  Take the age bands from the severity file rather than require that it matches up

johnlees:
This is a small change to `parameters.R`:
- Removes the age limits argument and takes it from fail
- Make the seeding argument correspondingly less painful to provide
- Allows for a user provided severity, but makes strict checks on column presence (with helpful error message)

Closes #4


richfitz:
Thanks - this moves things along nicely 😁

Issue #13, opened 2020-04-02
Remove support.R file


richfitz:
No longer used so we can remove this to clean up the parameters

PR #14, opened 2020-04-03
Fixing issues in the generate_parameters function

MJomaba:
Fixing issues in the generate_parameters function and changing default values to reflect latest model


johnlees:
@MJomaba I had a couple of these changes on the `model-fixes` branch, but I will discard that and merge it into here as well. I can fix the test failures for you too

johnlees:
I've fixed the testing errors. Also removed the `support.R` file so closes #13 

I updated two Rds files used in tests:
- The sampler test, I updated the log-likelihood and states, which are affected by the fixes to parameter value generation.
- The parameter test, where two of the proportions have been changed

Issue #7, opened 2020-04-02
Connect to processed data in ncov-outputs


richfitz:
This is a prerequisite for the rtm workflow; we need something where we are consuming data that is coming from the cleaned data tasks.

Requires at least #3 as we'll be doing a grid search, interacts with #4 if we'll start consuming severity data

PR #16, opened 2020-04-03
Simulate lockdown

johnlees:
This adds a main function (and tests) to run the particle filter on data from ncov outputs, as outlined by Marc: https://github.com/mrc-ide/sircovid-sandbox/tree/NHS-predictions

This also allows a time varying beta (new exported function for this). See `test-sircovid.R` for an example of how to run this yourself.

I think this address #7, but I'm not sure if it closes it yet. Any thoughts there?


richfitz:
To close #7, this needs to be done in ncov-outputs, which @OJWatson is most of the way through I think? Current work is here: https://github.com/ncov-ic/ncov-outputs/pull/171

So once this goes in, that task gets updated and simplified

OJWatson:
sorry missed that last night. will do .

R/parameters.R
```diff
@@ -1,3 +1,48 @@
...
+##' 
+##' @param beta_start Value of beta before intervention
+##' 
+##' @param start_date Start date for simualtions
+##' 
+##' @param reduction_start Start of reduction in beta
+##' 
+##' @param beta_reduction Factor by which beta is reduced
+##'   default is 0.238, as estimated by Jarvis et al
+##'   beta_reduced = beta - beta * (1 - beta_reduction)
```
MJomaba, line 14: simplifies to beta_reduced=beta_reduction? not sure I understand the line

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: sircovid
 Title: SIR Model for COVID-19
-Version: 0.2.3
+Version: 0.2.4
```
MJomaba, line 3: We might be going to 0.2.5 now. :)

R/parameters.R
```diff
@@ -1,3 +1,48 @@
...
+                          beta_reduction = 0.238, 
+                          reduction_period = 10) {
+  if (as.Date(start_date) > as.Date(reduction_start)) {
+    stop("Start date must be earlier than intervention date")
+  }
+  if (beta_reduction < 0 || beta_reduction > 1) {
+    stop("beta must decrease, and cannot be reduced below 0")
+  }
+  if (reduction_period > 100) {
+    message("Reduction period over 100 days - is this correct?")
```
MJomaba, line 32: I like this message, but we might reconsider in 3 months  📆

R/sircovid.R
```diff
@@ -1,7 +1,127 @@
...
+##' 
+##' @param obs_params List of parameters used for comparing
+##'   model to data in the particle filter
+##'   
+##' @param n_particles Number of particles  
+##' 
+##' @returns Results from particle filter
+##' 
+##' @export
+run_particle_filter <- function(case_data_file,
```
MJomaba, line 51: case_data_file is actually deaths? i think it is a bit confusing. I think at the moment it is not very generic, i.e. we will have soon to add the bed time series also. I think though we should stick to this at the moment to get this working. We can think of a more generic function later, and do a bit of brain storm. It would be good to have a bit of a brainstorm too as we come from different background and we want to encompass the statistic/mathematic logic but with the architecture/computer science robustness and flexibility

R/sircovid.R
```diff
@@ -1,7 +1,127 @@
...
+  # Merge mortality and admissions data, allowing NAs where there
+  # is no overlap
+  data <- merge(admissions_data, case_data, 
+                by.x="date", by.y="dateRep",
+                all.x=TRUE, all.y=TRUE)[,c("date", "deaths", "itu")]
+  if (nrow(data) < 1 || 
+      nrow(data) == nrow(admissions_data) + nrow(case_data) || 
+      ncol(data) != 3) {
+    stop("Merge of case and admissions data failed - do dates match?")
+  }
```
MJomaba, line 105: @OJWatson is reconstructing the times series in his function for sampling the grid, we might want to integrate the two approaches later.

R/parameters.R
johnlees: This refers to L40:
```
beta_slope <- beta_start * (1 - (1 - beta_reduction ) * (seq(0, reduction_period - 1) / (reduction_period - 1)))
```
Which for b=0.1, reduction - 0.234 over 10 days:
```
0.10000000 0.09148889 0.08297778 0.07446667 0.06595556 0.05744444 0.04893333 0.04042222 0.03191111 0.02340000
```
over 4 days:
```
0.10000000 0.07446667 0.04893333 0.02340000
```
which I think matches with your original

R/parameters.R
johnlees: Is that right? Would there potentially be a smooth reduction over such a long time?

R/sircovid.R
johnlees: I don't think it's too bad for a start, as it takes two files which just need to have date columns, one deaths and one itu (with arbitrary name). If this is expanded later to more files and column names, probably a more general `input_files`, `date_columns` and `output_columns` could be used, but don't know whether this is worth it now. What do you think?

Will change case to deaths if that makes it clearer - named cases because the downloaded file is called case data, though it also contains deaths (which I realise is what is actually being used)

R/sircovid.R
johnlees: https://github.com/mrc-ide/sircovid/blob/master/R/grid_search.R#L88-L98 could eventually be replaced by this function then?

MJomaba:
Hi @johnlees just testing the code on my machine; it runs and looks good.

PR #18, opened 2020-04-04
particle_filter: Adding in option save_particles = 'last' ...

lwhittles:
...to save a randomly selected particle at the final time point in the data, for use in the pmcmc filter

- save_particles = 'last' will also return the number of the particle chosen in the results, this is useful for checking the code but may be redundant otherwise so would be happy to remove.
- the existing save_particles options have been kept as TRUE/FALSE to keep backwards compatibility of code for people already using the package
- tests updated to check the new outputs


R/sampler.R
```diff
@@ -64,7 +72,7 @@ particle_filter <- function(data, model, compare, n_particles,
     step <- c(data$step_start[t], data$step_end[t])
     prev_state <- state
     state <- particle_run_model(state, step, model)
-    if (save_particles) {
+    if (save_particles == TRUE) {
```
richfitz, line 75: please revert this change

R/sampler.R
```diff
@@ -79,7 +87,7 @@ particle_filter <- function(data, model, compare, n_particles,
 
       kappa <- resample(weights$weights, "systematic")
       state <- state[, kappa]
-      if (save_particles) {
+      if (save_particles == TRUE) {
```
richfitz, line 90: please revert

R/sampler.R
```diff
@@ -96,11 +104,15 @@ particle_filter <- function(data, model, compare, n_particles,
   }
 
   ret <- list(log_likelihood = log_likelihood)
-  if (save_particles) {
+  if (save_particles == TRUE) {
```
richfitz, line 107: please revert

R/sampler.R
```diff
@@ -29,18 +34,21 @@ particle_filter <- function(data, model, compare, n_particles,
   if (n_particles < 2) {
     stop("At least two particles required")
   }
-  if (forecast_days > 0 && !save_particles) {
-    stop("forecasting only possible if particles are saved")
+  if (forecast_days > 0 && save_particles != TRUE) {
```
richfitz, line 37: please revert

R/sampler.R
```diff
@@ -14,8 +14,13 @@
 ##' @param forecast_days Number of days to forecast forward from end
 ##'   states.  Requires that \code{save_particles} is \code{TRUE}.
 ##'
-##' @param save_particles Logical, indicating if we save full particle
-##'   histories (this is slower).
+##' @param save_particles Either
```
richfitz, line 17: I don't love this interface.  Can you change this so that we add a new parameter:

```
save_sample_trajectory
```

or something, and if that's true create a new element in the return object

PR #19, opened 2020-04-06
Grid sample forecast

OJWatson:
Grid sample forecasting in and bump to 0.2.7 (the news version and Description was out of sync. Think we are on 0.2.6.


R/sample_grid.R
```diff
@@ -0,0 +1,109 @@
+##' Take a grid search produced by \code{\link{scan_beta_date}} and 
+##' sample \code{n_sample_pairs} from the parameter grid uses based
+##' on their probaility. For each parameter pair chosen, run particle
```
MJomaba, line 3: probability

R/sample_grid.R
```diff
@@ -0,0 +1,109 @@
...
+##' @return List. First element is an array of trajectories, in which
+##'  first dimension is time, second is parameters chosen when sampling from 
+##'  the grid and the third dimension is alist of inputs.
+##' 
+##' @export
+##' @import furrr
+##' @importFrom utils tail
+sample_grid_scan <- function(scan_results, 
+                             n_sample_pairs = 10, 
+                             n_particles = 100) {
```
MJomaba, line 24: n_trajectories would be more appropriate? if not I think it is confusing with the number of particles you run your particle filter with (which decide the "accuracy" of it)

R/sample_grid.R
```diff
@@ -0,0 +1,109 @@
...
+  ## Particle filter outputs
+  pf_run_outputs <- furrr::future_pmap(
+    param_grid,
+    function(beta, start_date) {
+      # Edit beta in parameters
+      model_params$beta_y <- beta
+      model_params$beta_t <- 0
+      
+      # Objects used by particle filter run
+      pf_data <- particle_filter_data(data, start_date, steps_per_day= 1/model_params$dt)
```
MJomaba, line 66: Shall we have an option to run the particle filter with a given number of particle? I guess here you have no choice other than default, you might want to increase (for more precision) or decrease (for speed) that number. We should have a parameter n_particle_for_filter. What we are running now would then be the default.

R/grid_search.R
```diff
@@ -138,3 +139,27 @@ plot.sircovid_scan <- function(x, ..., what = "likelihood") {
...
+##' @export
+plot.sample_grid_search <- function(x, ..., what = "ICU") {
+  
+  # what are we plotting
+  
+  if (what == "ICU") {
+    
+    index <- c(odin_index(x$inputs$model)$I_ICU) - 1L
+    ylab <- "ICU"
+    particles <- apply(x$trajectories[, index, ], c(1, 3), sum, na.rm = TRUE)
```
johnlees, line 151: Is this type-safe? See https://github.com/mrc-ide/sircovid-sandbox/pull/1#discussion_r402782849
i.e. will this definitely be numeric

R/grid_search.R
```diff
@@ -138,3 +139,27 @@ plot.sircovid_scan <- function(x, ..., what = "likelihood") {
...
+    
+  } else if(what == "Deaths") {
+    
+    index <- c(odin_index(x$inputs$model)$D) - 1L
+    ylab <- "Deaths"
+    particles <- apply(x$trajectories[, index, ], c(1, 3), sum, na.rm = TRUE)
+    plot_particles(particles, ylab = ylab)
+    points(as.Date(x$inputs$data$date), cumsum(x$inputs$data$deaths), pch = 19)
+    
+  }
```
johnlees, line 163: Can you add an `else` to give an error if an unimplemented `what` is passed

R/sample_grid.R
```diff
@@ -0,0 +1,109 @@
...
+##' @param scan_results Output of \code{\link{scan_beta_date}}.
+##' 
+##' @param n_sample_pairs Number of parameter pairs to be sampled. Integer.
+##'   Default = 10.
+##'   
+##' @param n_particles Number of particles. Positive Integer. Default = 100
+##'   
+##' @return List. First element is an array of trajectories, in which
+##'  first dimension is time, second is parameters chosen when sampling from 
+##'  the grid and the third dimension is alist of inputs.
```
johnlees, line 17: a list? Or function alist?

R/sample_grid.R
```diff
@@ -0,0 +1,109 @@
...
+  assert_pos_int(n_sample_pairs)
+  assert_pos_int(n_particles)
+  
+  # grab the pobability matrix
+  prob <- scan_results$renorm_mat_LL
+  nr <- nrow(prob)
+  nc <- ncol(prob)
+  
+  # construct what the x and y dimensions look like
+  x_grid <- matrix(scan_results$x, nrow = nr, ncol = nc)
```
johnlees, line 37: Could using `expand.grid()` be used here, which is the same method to generate this as in `sample_grid.R`

R/sample_grid.R
```diff
@@ -0,0 +1,109 @@
...
+  # recreate parameters for re running
+  param_grid <- data.frame("beta" = beta, "start_date" = dates)
+  model_params <- scan_results$inputs$model_params
+  pars_obs <- scan_results$inputs$pars_obs
+  data <- scan_results$inputs$data
+  
+  
+  # Multi-core futures with furrr (parallel purrr)
+  
+  ## Particle filter outputs
```
johnlees, line 57: This is identical to the code in `grid_search.R`. Can you make this into its own function, which is then called in each of these two places

R/sample_grid.R
```diff
@@ -0,0 +1,109 @@
...
+    x$states[,,sample(n_particles, 1)]
+  })
+  
+  
+  # collapse into an array of trajectories
+  # the trajectories are different lengths in terms of dates
+  # so we will fill the arrays with NAs where needed
+  num_rows <- unlist(lapply(traces, nrow))
+  max_rows <- max(num_rows)
+  seq_max <- seq_len(max_rows)
```
johnlees, line 86: I would just directly use this below rather than assigning to a new variable, I think it's clearer to read

johnlees: i.e. ` trajectories[tail(seq_len(max_rows), nrow(traces[[i]])), , i] <- traces[[i]]`

R/grid_search.R
OJWatson: yup good catch will do.

R/sample_grid.R
OJWatson: yup will add `\code{\link{list}}` to make it clearer

R/sample_grid.R
```diff
@@ -0,0 +1,109 @@
...
+  trajectories <- array(NA, 
+                        dim = c(max_rows, ncol(traces[[1]]), length(traces)),
+                        dimnames = list(rownames(traces[[which.max(num_rows)]]), NULL, NULL))
+  
+  # fill the tail of the array slice
+  for (i in seq_len(length(traces))){
+    trajectories[tail(seq_max, nrow(traces[[i]])), , i] <- traces[[i]]
```
johnlees, line 93: This is so that the end of the trajectories array is populated, and the start is padded with `NA` if it's shorter than the max. If that's correct, could you clarify slightly as a comment (e.g. note the output array size, and whether the NA padding is at the start or the end)

R/sampler.R
```diff
@@ -260,7 +260,7 @@ plot_particles <- function(particles, ylab) {
   ## Need to set plot up first to get the dates to render on the axis
   ## (matplot does not cope with this)
   dates <- as.Date(rownames(particles))
-  plot(dates, particles[, 1], type = "n", ylab = ylab, ylim = range(particles))
+  plot(dates, particles[, 1], type = "n", ylab = ylab, ylim = range(particles, na.rm = TRUE))
```
johnlees, line 263: Any other effects of potentially having `NA`s in output now? Tests look ok so I'm assuming it's fine

tests/testthat/test-sample_grid.R
```diff
@@ -0,0 +1,59 @@
...
+  data$deaths[tail(seq_len(nrow(data)), 2)] <- NA
+  
+  
+  # Parameters for run
+  min_beta <- 0.10
+  max_beta <- 0.18
+  beta_step <- 0.04
+  first_start_date <- "2020-01-29"
+  last_start_date <- "2020-02-14"
+  day_step <- 6
```
johnlees, line 33: Potentially use a lower number of steps if possible, this was a bit slow to test for me

R/sample_grid.R
OJWatson: Marc - i'm not sure I understand here. The n_particles parameter controls the number of particles being used in the particle filter at the moment. And n_sample_pairs will determine how many trajectories are run.

MJomaba: Sorry, my confusion, I got confused between particle_filter and particle_filter_data. Forget about the stupid comment!

R/sample_grid.R
OJWatson: done

R/sample_grid.R
OJWatson: Wil resolve by adding to documentation to show that n_sample_pairs controls the number of trajectories.

R/sample_grid.R
OJWatson: Resolved with more documentation

R/grid_search.R
OJWatson: done

R/sample_grid.R
OJWatson: resolved with documentation

R/sample_grid.R
OJWatson: This is slightly different to how `expand.grid` is used in `grid_search` as trying to get the grid of beta and start values that correspond to the z axis matrix:

```
x_grid
     [,1] [,2]
[1,] 0.10 0.10
[2,] 0.18 0.18
```

Will leave as is if that makes sense

R/sample_grid.R
OJWatson: Yeah we could do that actually as i just stole it. Will do that

R/sample_grid.R
johnlees: Ah right, I misunderstood. Can you just clarify this example output in a comment?

R/sample_grid.R
```diff
@@ -0,0 +1,109 @@
...
+      model_run <- sircovid(params = model_params)
+      compare_func <- compare_icu(model_run, pars_obs, pf_data)
+      
+      X <- particle_filter(pf_data, model_run, compare_func, 
+                           n_particles = n_particles, save_particles = TRUE)
+    }
+  )
+  
+  # sample 1 particle from each run
+  traces <- lapply(pf_run_outputs, function(x) {
```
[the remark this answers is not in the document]
OJWatson: yes as the matrix returned is not always the same size

R/sampler.R
OJWatson: yeah that was my hope too. can't see why it would impact the ylim setting as previously the trajacetories would never have NAs in as they are straight from odin

tests/testthat/test-sample_grid.R
OJWatson: done and changed the n_particles down as that will speed this up too

R/sample_grid.R
OJWatson: added comment

R/sample_grid.R
OJWatson: interesting ... I only rewrote it out like this as i wrote it in full and thought that looks really bloated and not clear at all. Is it alright to leave this as is @johnlees? Mainly being as i want `seq_max` later on and it keeps it all under 80 character line width :)

R/sample_grid.R
OJWatson: yup done

Issue #20, opened 2020-04-06
Add pMCMC sampler


johnlees:
@lwhittles has done this here: https://github.com/mrc-ide/sircovid-sandbox/pull/1

I think it's ready for integration into this package

lwhittles:
Great, thanks John! Is there a clever way to do this, or should I just open a new branch in sircovid and copy the code across in to the relevant directories?

johnlees:
I think there are some ways which would preserve the commit history, but they look like they are more complex than it's worth. So I'd just copy over the two new .R files, commit to a new branch, and make a PR here.
With the previous code moved from here we did this too, so it should be ok.

lwhittles:
I have branched off sircovid https://github.com/mrc-ide/sircovid/tree/pmcmc
and integrated the pmcmc and testing scripts from the sandbox https://github.com/mrc-ide/sircovid-sandbox/tree/pmcmc_dev. Just running the build tests now and will create a PR shortly

PR #25, opened 2020-04-06
Pmcmc

lwhittles:
Branched off sircovid https://github.com/mrc-ide/sircovid/tree/pmcmc
and integrated the pmcmc and testing scripts from the sandbox https://github.com/mrc-ide/sircovid-sandbox/tree/pmcmc_dev.


johnlees:
Closes #20 when merged

R/pmcmc.R
```diff
@@ -0,0 +1,428 @@
...
+  if(is.null(pars_obs)) {
+    pars_obs <- list(phi_ICU = 0.95,
+                     k_ICU = 2,
+                     phi_death = 926 / 1019,
+                     k_death = 2,
+                     exp_noise = 1e6
+    )
+  }
+  
+  if(is.null(pars_init)) {
```
johnlees, line 124: Can these defaults for the beta ranges instead be provided as default function arguments above, rather than having everything as `NULL`. I think this format makes sense for the more complex arguments/functionally generated arguments above, but for this simpler case having them above makes it a bit easier to see where they are if changed in the future. I think doing this with `pars_obs` would also make sense, as that's more consistent with other parts of the code.

If you have time, quick checks that these are in a reasonable range would be useful, but I am happy to do this if helpful

R/pmcmc.R
```diff
@@ -0,0 +1,428 @@
...
+                            start_date = start_date, 
+                            steps_per_day = steps_per_day)
+  compare <- compare_icu(mod, pars_obs, d)
+  X <- particle_filter(data = d, 
+                       model = mod, 
+                       compare = compare, 
+                       n_particles = n_particles)
+  # This is currently a placeholder for when the particle filter function in the sircovid package
+  # function is changed to allow output of a sample state 
+  X$sample_state <- rep(1, 238)
```
johnlees, line 426: I think your change to allow this is now merged, can this be updated to use it?

R/pmcmc.R
```diff
@@ -0,0 +1,428 @@
...
+  
+  states <- matrix(data = NA,
+                   nrow = n_mcmc, 
+                   ncol = length(ss_init))
+  
+  
+  ## create function to reflect proposal boundaries at pars_min and pars_max
+  # this ensures the proposal is symetrical and we can simplify the MH step
+  
+  reflect_proposal <- function(x, floor, cap) {
```
johnlees, line 319: Make these lambdas/functionals into their own internal functions, as with `calc_loglikelihood`. The main function is quite long and this would make it easier to read

R/pmcmc.R
```diff
@@ -0,0 +1,428 @@
...
+  
+  # extract loglikelihhod estimate and sample state
+  # calculate posterior
+  curr_ll <- p_filter_est$log_likelihood
+  curr_lpost <- curr_lprior + curr_ll
+  
+  curr_ss <- p_filter_est$sample_state
+  
+
+  ## combine into a vector to add to the results at the end
```
johnlees, line 295: Can these just be added into `res` and `states` immediately? If you set `nrow = n_mcmc + 1`

R/pmcmc.R
```diff
@@ -0,0 +1,428 @@
...
+  }
+  # end progress bar
+  close(pb)
+  
+  
+  # include starting conditions in output
+  res <- rbind(out_init, res, deparse.level = 0)
+  res <- as.data.frame(res)
+  
+  tmp <- coda::as.mcmc(res)
```
johnlees, line 388: Change `tmp` to `coda_res` or similar

johnlees: Actually just remove this, if it can be added above

R/pmcmc.R
```diff
@@ -0,0 +1,428 @@
...
+  # this ensures the proposal is symetrical and we can simplify the MH step
+  
+  reflect_proposal <- function(x, floor, cap) {
+    interval <- cap - floor
+    abs((x + interval - floor) %% (2 * interval) - interval) + floor
+  }
+  
+  
+   # this can easily be adapted to being multivariate normal with covariance to improve mixing
+  propose_jump <- function(pars) {
```
johnlees, line 326: This would also be better as its own internal function

R/pmcmc.R
```diff
@@ -0,0 +1,428 @@
...
+    
+    ## proposed jumps are normal with mean pars and sd as input for parameter
+    tmp <- rnorm(n = n_pars, mean = pars, sd = pars_sd)
```
johnlees, line 329: Rename `tmp` to `jumps`

R/pmcmc.R
lwhittles: default function arguments passed as requested and pushed
leaving the conversation unresolved for now so I remember to add reasonable range checks

R/pmcmc.R
```diff
@@ -0,0 +1,428 @@
...
+                            n_particles = n_particles
+    ) 
+    X
+  }
+  
+  
+  
+  # convert the current parameters into format easier for mcmc to deal with
+  curr_pars <- unlist(pars_init)
+  day_zero <- as.Date(data$date[1])
```
johnlees, line 222: How does this relate to the first data date, and the model start date? I guess it's day zero in the same sense as the data, but the model will run from prior to this. I think this is fine as it's consistent between this and the particle sampler, but might be useful to use a consistent variable name (see the different start date types in `run_particle_filter()`)

R/pmcmc.R
lwhittles: change made and pushed

tests/testthat/test-pmcmc.R
```diff
@@ -0,0 +1,337 @@
+context("pmcmc")
+
+data <- read.csv(system.file("extdata/example.csv",
```
johnlees, line 3: Use `sircovid_file()` instead of `system.file`

johnlees: Put this in each function, rather than using a global

tests/testthat/test-pmcmc.R
```diff
@@ -0,0 +1,337 @@
...
+  X <- pmcmc(
+    data = data, 
+    n_mcmc = n_mcmc
+  )
+  
+  
+ expect_is(X, 'list')
+ expect_setequal(names(X), c('inputs', 'results', 'states', 'acceptance_rate', 'ess'))
+ expect_equal(dim(X$results), c(n_mcmc + 1L, 5))
+ expect_equal(dim(X$states), c(n_mcmc + 1L, 238))
```
johnlees, line 20: It might be useful to save the run with this seed as `pmcmc.Rds`, load it back in and check that it's identical. See "Parameters are generated as before" in `test-parameters.R` and L47-55 in `test-sampler.R`

R/pmcmc.R
lwhittles: done and pushed

R/pmcmc.R
lwhittles: I have taken the reflect_proposal and propose_parameters functions out of the pmcmc sampler and created internal functions. I intend to write tests for these functions so am leaving the conversation unresolved for now.

tests/testthat/test-pmcmc.R
lwhittles: done and pushed

R/pmcmc.R
lwhittles: done and pushed

R/pmcmc.R
lwhittles: done and pushed - still need to write checks

R/pmcmc.R
lwhittles: I have made a couple of simplifying changes that should improve readability, including converting the data$date vector into Date class at the beginning of the code, rather than multiple times throughout, and removing the need for creating a 'day_zero' object

R/pmcmc.R
lwhittles: done and pushed

tests/testthat/test-pmcmc.R
lwhittles: added check and pushed

R/pmcmc.R
lwhittles: added in reasonable range checks and pushed

R/pmcmc.R
lwhittles: added in tests

PR #32, opened 2020-04-06
fixing beta_t so it matches the correct time scale (time-steps not days)

edknock:
Before the values in beta_t were on a time-scale in terms of days, but this is used in the model on a time-scale of time-steps of length dt, so beta_t needs to be on that time-scale.

Hence scale beta_t by 1/dt.

### DIFF
--- a/tests/testthat/test-sircovid.R
+++ b/tests/testthat/test-sircovid.R
@@ -19,7 +19,7 @@ test_that("Can be run on real data", {
                                                    exp_noise = 1e6),
                                  n_particles = 1000)
   # No check of correctness  
-  expect_equal(results$log_likelihood, -14074.79, tolerance=1e-3)
+  expect_equal(results$log_likelihood, -269.9478, tolerance=1e-3)
 })
 
 test_that("Poor formatting of real data errors", {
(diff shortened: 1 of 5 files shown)


edknock:
Yes have looked at results and am happy with them.

johnlees:
@richfitz this is actually an easier change than first thought, as most of the reference runs didn't use a time varying beta. So just needed to add a scaling to dates in the tests too.

richfitz:
Thanks @johnlees - I'd not had a chance to properly look through the failures :)

tests/testthat/test-sircovid.R
```diff
@@ -19,7 +19,7 @@ test_that("Can be run on real data", {
                                                    exp_noise = 1e6),
                                  n_particles = 1000)
   # No check of correctness  
-  expect_equal(results$log_likelihood, -14074.79, tolerance=1e-3)
+  expect_equal(results$log_likelihood, -269.9478, tolerance=1e-3)
```
richfitz, line 22: that is quite an improvement...

richfitz:
Thanks @johnlees - I'd not had a chance to properly look through the failures :)

PR #34, opened 2020-04-07
Refactor pf sampler runs, improve memory use

### LINKED ISSUE #29: Use the beta_date_particle_filter function rather than repeating
Reused in `run_particle_filter()`, which should process data, and call this, and in the MCMC

### LINKED ISSUE #31: Memory Issues
Rich just sent me the following issue when testing the grid search through the new orderly task:

```
[17:34] Fitzjohn, Richard G
    
^[[BError in getGlobalsAndPackages(expr, envir = envir, tweak = tweakExpression,  : 
  The total size of the 7 globals that need to be exported for the future expression (‘{; ...future.f.env <- environment(...future.f); if (!is.null(...future.f.env$`~`)) {; if (is_bad_rlang_tilde(...future.f.env$`~`)) {; ...future.f.env$`~` <- base::`~`; }; ...; .out; }); }’) is 6.31 GiB. This exceeds the maximum allowed size of 500.00 MiB (option 'future.globals.maxSize'). The three largest globals are ‘...future.x_ii’ (6.31 GiB of class ‘list’), ‘is_bad_rlang_tilde’ (15.05 KiB of class ‘function’) and ‘...future.map’ (6.61 KiB of class ‘function’).
Calls: tryCatch ... SequentialFuture -> UniprocessFuture -> getGlobalsAndPackages
```

We are clearly hitting some memory issues - I tried running it locally and it gets very close to 16Gb. Something is clearly an issue in future or somewhere. Posting here till we figure out what

johnlees:
I've separated the data generation and fitting steps into two exported functions, which I think makes more sense conceptually. This also allows the `run_particle_filter()` function to be more effectively reused (e.g. in the grid search, grid samples and eventually the MCMC) reducing some redundancy.

I hope this should help with the memory issues as the entire trajectories of all particles on the grid are no longer being returned, but would appreciate a check with the orderly run that was failing yesterday.

When merged closes #29  and closes #31


OJWatson:
Hi @johnlees  - I haven't got time to do a full review. But have tested locally and this fixes the memory issue at #31 both within `sample_grid_scan` and `scan_beta_date`

johnlees:
@OJWatson that's great, thank you! That's all I wanted for now

PR #37, opened 2020-04-08
plot daily deaths not cumulative

lwhittles:
Plotting daily deaths rather than cumulative - note that the plotted deaths data are now quite far outside of the projections produced by the https://github.com/ncov-ic/ncov-outputs/tree/rtm_itu_deaths_for_sircovid/src/rtm_inference_sircovid task

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: sircovid
 Title: SIR Model for COVID-19
-Version: 0.2.11
+Version: 0.2.12
 Authors@R:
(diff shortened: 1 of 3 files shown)


edknock:
I have made a further edit to @lwhittles's, plots daily death counts for the particles now as well as the data. Plotted deaths data are now generally within the projections produced by the https://github.com/ncov-ic/ncov-outputs/tree/rtm_itu_deaths_for_sircovid/src/rtm_inference_sircovid task.

Added @johnlees as an additional reviewer.

PR #38, opened 2020-04-09
adding save_end_states option to particle_filter

edknock:
When we run the particle filter within the SMC^2 algorithm, need to output all the end states (along with the log-likelihood). Can extract this with the save_particles option, but the save_end_states option is faster and saves on memory.

### DIFF
--- a/R/sampler.R
+++ b/R/sampler.R
@@ -21,7 +21,7 @@
 ##'  single particle, chosen at random, at the final time point for which 
 ##'  we have data
 ##'  
-##'  @param save_end_states Logical, indicating whether we should save all
+##' @param save_end_states Logical, indicating whether we should save all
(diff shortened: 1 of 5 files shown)


R/sampler.R
```diff
@@ -20,12 +20,15 @@
...
+##'  
+##'  @param save_end_states Logical, indicating whether we should save all
+##'  particles at the final time point for which we have data
 ##'   
 ##' @export
 ##' 
 particle_filter <- function(data, model, compare, n_particles,
                             forecast_days = 0, save_particles = FALSE, 
-                            save_sample_state = FALSE) {
+                            save_sample_state = FALSE, save_end_states = FALSE) {
```
johnlees, line 31: As discussed, `stop()` if `save_particles && save_end_states`

PR #46, opened 2020-04-14
Flexible generate beta

### LINKED ISSUE #21: Adapt the generate_beta function to take two parameters
Allow the function to take two parameters instead of one, if the second parameter is not specified, used the current reduction.

johnlees:
Allow passing of generate beta as a function, should it ever be changed

Closes #21


johnlees:
I'd left this hanging because I was going to do some other issues at the same time, it's not the most exciting thing but best put it in before we sort out the multiple models I reckon

R/grid_search.R
```diff
@@ -38,7 +42,10 @@
 scan_beta_date <- function(
   min_beta, 
   max_beta, 
-  beta_step, 
+  beta_step,
+  generate_beta_func = function(beta_start, start_date)
```
richfitz, line 46: this is the same as `generate_beta_func = generate_beta` I believe

R/pmcmc.R
```diff
@@ -86,6 +90,9 @@ pmcmc <- function(data,
                   n_mcmc, 
                   model_params = NULL,
                   model_start_date = "2020-02-02",
+                  generate_beta_func = function(beta_start, start_date)
```
richfitz, line 93: also here, can be done more simply.  However if we're repeating it everywhere should we look at lumping this into the generic "model" object perhaps?

johnlees:
Thanks, I've simplified those calls. I'd merged the model change in 5f0122e but have updated again.

Good point about the model objects, perhaps we do just want to give them their own beta generating function when we initialise them. I'm not sure if we should just merge this for now and I can think about that now I'm sorting the rest out?

PR #49, opened 2020-04-15
update on parameter function

MJomaba:
update on parameter function to take fitted values from COCIN/CHESS hospital model


R/parameters.R
```diff
@@ -411,6 +412,26 @@ generate_parameters_new_hospital_model <- function(
   I0_asympt[seed_idx,1,N_trans_classes] <- infection_seeding$values
   S0[seed_idx] <- S0[seed_idx] - infection_seeding$values
   
+  if (hospital_fitted){
+  
+    # Read file with fitted parameters (from Bob Verity's hospital model)
+    fitted_parameters <- readRDS(file = "inst/extdata/fitted_parameters.rds")
```
richfitz, line 418: how is this file generated? where is it generated? how is it updated?

MJomaba: This file is generated from a spreadsheet, coming from Bob's model and processed by Neil. I have no idea of how it will move forward, but ideally we would like to get this information directly from Bob's ultimately. The fitting is to transform a cum inverse gamma distribution into an Erlang one using KL divergence. The code for this is on the sandbox.
https://github.com/mrc-ide/sircovid-sandbox/pull/2

johnlees:
Can I suggest that we keep the previous severity file too, to avoid changing lots of tests? Perhaps we can add them with a time/datestamp from now on?

I also notice in the new file that there are a few spaces in the 'cells' of the CSV (visible in the diff) which should probably be removed

johnlees:
Ah, I see that you've done that already @MJomaba, sorry was missing that part of the diff! To fix the tests it may be easiest to follow the time/datestamp idea for `severity.csv` updates, and keep the current one named as-is.

The new field with the space I noticed is `Unadjusted_IFR ,`

R/parameters.R
```diff
@@ -382,7 +383,34 @@ generate_parameters_new_hospital_model <- function(
...
+  #this is calculated externally at the moment
+    if (hospital_fitted){
+    
+    fitted_parameter_file <- sircovid_file("extdata/fitted_parameters.rds")
+    
+    # Set up severity file into table
+    
+    
+    # Read file with fitted parameters (from Bob Verity's hospital model)
+    fitted_parameters <- readRDS(file = fitted_parameter_file)
```
johnlees, line 397: I'd prefer if we made this into a CSV file and read it in like we do with the severity data. If we end up merging #50 first this will need to be changed a bit too, as these go in the models rather than the parameters

PR #50, opened 2020-04-15
Make alternative models into objects

### LINKED ISSUE #21: Adapt the generate_beta function to take two parameters

### LINKED ISSUE #42: Generalise addition of new models
Each new model will have:
- New odin code
- A name (also the class name)
- Comparison function (inherited from compare_icu)
- Parameter generating function (inherited from generate_parameters)
- Plotting functions

I think if consumers of these models (mcmc, grid_search) are passed a model object and do not have to worry about which specific model is used, that would be ideal.

johnlees:
This tries to make adding and using new/different models a bit easier, and also tries to reduce the use of `model` to refer to many different things. Each model is its own object with a `odin_model`, `generate_beta_func` and `compare_model` functions. Progression groups have been moved from parameters to model (they felt more appropriate here) as have gammas (though we might want to change this considering #49).

I've removed redundancy in `compare.R` and `parameters.R`. I couldn't figure out a neat way to do this with inheritance/polymorphism – I considered having a `generate_parameters()` you'd call on a model, but this was difficult to avoid repeating code, and having the return match exactly with what the odin model expects. It also seemed clearer to write out all of the compartment initialisations each time for each model, so you can see what is expected. 

So adding further models consists of:
- Writing odin code
- Defining a new model class in `models.R`
- Altering `generate_parameters()` and `compare_output()` if necessary, using the class name

Questions:
- A bit I'm still unhappy with is l164 of `parameters.R` where unused parameters are set to `NULL` manually. Is there a way to get these out of the odin model so they can be removed in a generic fashion? (leaving them in produces warnings.
- The pMCMC results tests fails, giving different output. I'm not immediately sure why when all the others work, but it could be that the defaults to some functions have changed.

Closes #42, closes #21


richfitz:
> A bit I'm still unhappy with is l164 of parameters.R where unused parameters are set to NULL manually. Is there a way to get these out of the odin model so they can be removed in a generic fashion? (leaving them in produces warnings.

You use the `unused_user_action` argument to the generator to control this behaviour - see https://mrc-ide.github.io/odin/reference/odin.html#user-parameters for possible options (you probably want `unused_user_action = "ignore"`)

PR #52, opened 2020-04-16
adding some plots for the new model with ICU, death and general

MJomaba:


johnlees:
I think we should merge this in, I'll merge it to #50 as this actually tests the new model, and I can add an appropriate plotting test to increase the coverage here

johnlees:
Actually maybe I'll just merge it in from here, to avoid fiddling with the release version, extra CI etc

PR #53, opened 2020-04-16
Tests for hospital model

edknock:
Added tests for hospital model, similar to tests for basic model.


tests/testthat/test-hospital-model.R
```diff
@@ -0,0 +1,265 @@
...
+  sircovid_model <- hospital_model()
+  pars_model <- generate_parameters(sircovid_model = sircovid_model, beta = 0.126)
+  pars_model$p_recov_ILI[]<-1
+  mod <- sircovid_model$odin_model(user = pars_model)
+  t_max <- 150
+  t <- seq(from = 1, to = t_max)
+  check_cases <- FALSE
+  max_iter <- 10
+  iter <- 0
+  while (!check_cases && iter<= max_iter){
```
johnlees, line 103: Can you add a comment to note why the loops are being used.
Also I think putting an `expect_true(any(results$I_ILI[,,,] > 0))` after the loop would make sense, to make sure if it exits after max_iter the correct failure is diagnosed

edknock: Have added comments about why the loops are being used. Added expect_true statements like you suggested after every while loop. Also switched p_death_hosp to p_death_hosp_D in line with the changes in 0.2.18.

PR #60, opened 2020-04-17
-removing model_start_date argument from pmcmc() as it duplicates par…

lwhittles:
…s_init$start_date

- moving pmcmc argument tests up, so that errors are caught before generate_parameters() is called
- add new test to make sure that pars_init are being passed to the mcmc pmcmc
- update docs
-update pmcmc ref tests to be bug-free


R/pmcmc.R
```diff
@@ -362,9 +361,9 @@ pmcmc <- function(data,
 }
 
 
-calc_loglikelihood <- function(pars, data, sircovid_model, model_params, model_start_date,
+calc_loglikelihood <- function(pars, data, sircovid_model, model_params,
                                steps_per_day, pars_obs, n_particles) {
-  start_date <- as.Date(pars[['start_date']], origin=model_start_date)
+  start_date <- as.Date(-pars[['start_date']], origin=data$date[1])
```
johnlees, line 366: This is the key change, looks like it makes sense in the debugger.

Can you just add a quick comment to explain that start_date is an int which we subtract from the first observed date in the data? Then good to merge

PR #64, opened 2020-04-21
multivariate normal proposals

lwhittles:
- added in mvtnorm dependency
- pars_sd input is now cov_mat
- new test
- bumped version number


R/pmcmc.R
```diff
@@ -109,8 +109,22 @@ pmcmc <- function(data,
                   steps_per_day = 4) {
   
   # test pars_init input
-  correct_format <- function(pars) {
-    is.list(pars) & setequal(names(pars), c('beta', 'start_date'))
+  correct_format <- function(pars, input = "") {
```
johnlees, line 112: I would suggest splitting this in two, as this new form doesn't share any of the code. You could leave the simple format of `correct_format()` you had previously as 'correct_format_vec()`, and add `correct_format_mat()` for the new check.

R/pmcmc.R
```diff
@@ -99,8 +100,7 @@ pmcmc <- function(data,
                                   'start_date' = 0),
                   pars_max = list('beta' = 1, 
                                   'start_date' = 1e6),
-                  pars_sd = list('beta' = 0.001, 
-                                 'start_date' = 0.5),
+                  cov_mat = c('beta' = 0.001^2, 'start_date' = 0.5^2),
```
johnlees, line 103: I think a form which shows this as a matrix (even with zeros) would be easier to modify for a user.
```
cov_mat = matrix(c(0.001^2, 0,
                     0, 0.5^2), 
```
would be equivalent?

tests/testthat/test-pmcmc.R
```diff
@@ -66,14 +73,36 @@ test_that("pmcmc runs without error", {
...
+X2 <- pmcmc(
+  data = data, 
+  n_mcmc = n_mcmc,
+  sircovid_model = sircovid_model,
+  pars_obs = list(phi_ICU = 0.95,
+                  k_ICU = 2,
+                  phi_death = 926 / 1019,
+                  k_death = 2,
+                  exp_noise = 1e6),
+  cov_mat = matrix(c(0.001^2, 0,
```
johnlees, line 89: Can you add a test which has non-zero covariance between proposals just to check it runs. I don't think there's any  need to check exact run results, but maybe that you get different results from `X2` with the same seed would be sensible

PR #66, opened 2020-04-21
Grid updates

### LINKED ISSUE #54: Put a time tag on forecast pictures
and potentially on other outputs?

### LINKED ISSUE #63: Add support for priors in the scan + sample functions on a grid
this can be easily done by calculating a e.g. gamma prior on the beta. The log of the betas can be easily added to the grid of the log-likelihood. The calculation of the probability and the sampling remains the same. A result for log prior matrix will be added in the scan results object, and a posterior matrix, the sampling will be done from the posterior rather than the log likelihood.

johnlees:
Mostly moves some of the stuff currently in ncov-outputs that should be in the package. I can follow up with a PR there to remove some of the code no longer needed

The priors were done as in outputs – a gamma prior on beta applied to the grid results. Let me know if this should change/be generalised.

Closes #54, closes #63


R/grid_search.R
```diff
@@ -122,18 +151,18 @@ scan_beta_date <- function(
...
   } else if (what == "probability") {
     graphics::image(x=x$x, y=x$y, z=x$renorm_mat_LL,
-                    xlab="beta", ylab="Start date", main = "Probability")
+                    xlab="beta", ylab="Start date", main = paste0("Probability (run ", date ,")"))
   }
 }
 
 ##' @export
-plot.sample_grid_search <- function(x, ..., what = "ICU") {
+plot.sample_grid_search <- function(x, ..., what = "ICU", date = as.character(Sys.Date())) {
```
richfitz, line 165: I think that we should be explicit about date throughout or we will get into trouble where the date of the *data* is not today's date.  That will happen fairly frequently (impacts at least the other plotting function too).  I spent some time removing this from a different piece of work unrelated to this workstream...

johnlees: Thanks, have changed it so that it can be set manually if wanted

Issue #62, opened 2020-04-18
Create a check for a grid_scan result


MJomaba:
in particular check that the border as probability near 0, if not generate a warning (useful when we are running automatically the grid search)

PR #70, opened 2020-04-22
Add a warning when edges of probability matrix stray away from 0

sangeetabhatia03:
Also added test for warning. This is intended to address issue #62. 
I had intended to make this generalisable so that it runs without modifications when a 3D grid is in place, but this is probably over-kill at this point as the code above it (that build a matrix) will need to edited anyway.

### DIFF
--- a/R/grid_search.R
+++ b/R/grid_search.R
@@ -131,11 +131,8 @@ scan_beta_date <- function(
   ## Check if the edges of the matrix in each dimension are close
   ## enough to 0.
   ## The first and last rows and the first and last columns
-  close_enough <- (all(prob_matrix[1, ] < tolerance) &
-    all(prob_matrix[, 1] < tolerance) &
-    all(prob_matrix[nrow(prob_matrix), ] < tolerance) &
-    all(prob_matrix[, ncol(prob_matrix)] < tolerance)
-  )
+  ## Or in case of multidimensional arrays, edges in each dimension.
+  close_enough <- zero_boundary(prob_matrix, tolerance = tolerance)
 
   if (!close_enough) {
     warning("Edges of the probability matrix are not close enough to 0.")
--- a/R/grid_search.R
+++ b/R/grid_search.R
@@ -1,24 +1,24 @@
 ##' Run a grid search of the particle filter over beta and start date.
-##' This is parallelised, first run \code{plan(multiprocess)} to set 
+##' This is parallelised, first run \code{plan(multiprocess)} to set
 ##' this up.
-##' 
+##'
 ##' @title Grid search of beta and start date
-##' 
+##'
 ##' @param min_beta Minimum value of beta in the search
-##' 
+##'
 ##' @param max_beta Maximum value of beta in the search
-##' 
+##'
 ##' @param beta_step Step to increment beta between min and max
-##' 
+##'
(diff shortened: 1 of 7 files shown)


R/grid_search.R
```diff
@@ -91,147 +94,175 @@ scan_beta_date <- function(
...
+  )
 
   # Exponentiate elements and normalise to 1 to get probabilities
   prob_matrix <- exp(mat_log_ll)
-  renorm_mat_LL <- prob_matrix/sum(prob_matrix)
-  
+  ## Check if the edges of the matrix in each dimension are close
+  ## enough to 0.
+  ## The first and last rows and the first and last columns
+  close_enough <- (all(prob_matrix[1, ] < tolerance) &
```
johnlees, line 134: I think it's worth generalising this, as although I'm not sure if we will extend past 2D these are the cases where this check is most useful as it becomes difficult to visualise the results. I would move this to a function that looks something like this:
```
# from https://stackoverflow.com/a/14502298
index_array <- function(x, dim, value, drop = FALSE) { 
  # Create list representing arguments supplied to [
  # bquote() creates an object corresponding to a missing argument
  indices <- rep(list(bquote()), length(dim(x)))
  indices[[dim]] <- value

  # Generate the call to [
  call <- as.call(c(
    list(as.name("["), quote(x)),
    indices,
    list(drop = drop)))
  # Print it, just to make it easier to see what's going on
  print(call)

  # Finally, evaluate it
  eval(call)
}

zero_boundary <- function(array, tolerance) {
  if (class(array) == "matrix") {
    close_enough <- (all(prob_matrix[1, ] < tolerance) &
    all(prob_matrix[, 1] < tolerance) &
    all(prob_matrix[nrow(prob_matrix), ] < tolerance) &
    all(prob_matrix[, ncol(prob_matrix)] < tolerance)
  } else {
    close_enough <- 1
    for (dimension in 1:dim(array)) {
      close_enough <- (all(index_array(array, dimension, 1) < tolerance &
       all(index_array(array, dimension, dim(array)[dimension]) < tolerance)
    }
  }
 close_enough
}

File: tests/testthat/test-grid_search.R [hunk 2]
```diff
@@ -46,40 +46,63 @@ test_that("Small grid search works with new model", {
...
+    beta_step = beta_step,
+    first_start_date = first_start_date,
+    last_start_date = last_start_date,
     day_step = day_step,
     data = data,
     sircovid_model = hospital_model(),
     scale_prior = 0.003541667,
-    shape_prior = 36)
+    shape_prior = 36,
+    ## 292 0s before a non-zero number when I run stuff.
```
johnlees, line 84: A better test might instead be to move the search boundaries so they are cutting through the probability, as this could be getting pretty close to machine precision. So perhaps try a min_beta of 0.17 (or look at the plot output from this to see where you could put a plot boundary to cut through the main mass)

tests/testthat/test-grid_search.R
```diff
@@ -46,40 +46,63 @@ test_that("Small grid search works with new model", {
...
+    ## Hence I am setting tolerance to be very high.
+    ## In practice, we migh need less.
+    tolerance = 1e-300
+    )
```
johnlees, line 88: Could you add the expected warning message

sangeetabhatia03:
Thanks for the suggestions John. I have generalised, this should work fine now.

PR #72, opened 2020-04-23
Introducing variable seeding

edknock:
Adding a function into sircovid_model that allows for variable seeding of the epidemic within the particle filter (variable both in terms of when the epidemic is seeded and what state it is seeded with). If pars_seeding is NULL then the particle filter should work exactly as before.

More generally, start_date now acts as an earliest possible seeding date.


R/sampler.R
```diff
@@ -249,3 +248,51 @@ scale_log_weights <- function(log_weights) {
   }
   list(weights = weights, average = average)
 }
+
+
+pre_data_run <- function(model, data, seeding_func,
```
johnlees, line 253: Should this also take a `save_particles` option, and return `particles = NULL` if true? Just thinking about memory use

R/sampler.R
```diff
@@ -49,28 +51,25 @@ particle_filter <- function(data, model, compare, n_particles,
...
-      state_with_history[, i_state, ]
-    state <- state_with_history[length(step), i_state, , drop = TRUE]
   } else {
     particles <- NULL
-    step <- c(data$step_start[[1L]], data$step_end[[1L]])
-    state <- model$run(step = step, use_names = FALSE, replicate = n_particles,
-                       return_minimal = TRUE)[, 1, , drop = TRUE]
   }
 
+  X <- pre_data_run(model = model,
```
johnlees, line 64: I find it a bit counter-intuitive that this takes the full `data` as a parameter. I think it might be better just to pass the parts of it that are needed (step and end).

R/sampler.R
johnlees: Actually I see that you have effectively done this already. I think it would be clearer and more consistent to make this explicit rather than checking `is.null`

R/sampler.R
```diff
@@ -49,28 +51,25 @@ particle_filter <- function(data, model, compare, n_particles,
   }
 
   i_state <- seq_along(model$initial()) + 1L
-
-  ## Special treatment for the burn-in phase; later we might use this
-  ## same approach for skipping steps though.
-  if (save_particles) {
+  
+  if (save_particles){
```
johnlees, line 55: Remove this section (see comments below)

R/sampler.R
```diff
@@ -249,3 +248,51 @@ scale_log_weights <- function(log_weights) {
...
+pre_data_run <- function(model, data, seeding_func,
+                         n_particles, i_state, particles){
+  
+  steps_per_day <- attr(data, "steps_per_day")
+  
+  state <- array(NA_real_,c(length(i_state),n_particles))
+  
+  seeding <- seeding_func(n_particles)
+  
+  if (!is.null(particles)) {
```
johnlees, line 262: In fact, I think you should get rid of the section in `particle_filter()` which sets up `particles` and move the array initialisation here. Get rid of the `particles` input to this function, add `save_particles` and return either `NULL` or the array depending on whether true or false

R/sampler.R
```diff
@@ -49,28 +51,25 @@ particle_filter <- function(data, model, compare, n_particles,
...
-    state <- model$run(step = step, use_names = FALSE, replicate = n_particles,
-                       return_minimal = TRUE)[, 1, , drop = TRUE]
   }
 
+  X <- pre_data_run(model = model,
+                    data = data,
+                    seeding_func = seeding_func, 
+                    n_particles = n_particles,
+                    i_state = i_state,
+                    particles = particles)
```
johnlees, line 69: Does this need to check that the first date in data is after the time being run up to, or will it work in either case?

R/seeding.R
```diff
@@ -0,0 +1,48 @@
...
+##' @param data The data, must be constructed with
+##'   \code{particle_filter_data}
+##'
+##' @param pars_seeding Parameters for seeding
+##' 
+##' @importFrom stats rpois
+##' 
+##' @export
+seeding_function <- function(model,
+                             data,
```
johnlees, line 16: Again, this would probably be better with the two parameters from data rather than the whole object

R/seeding.R
```diff
@@ -0,0 +1,48 @@
+##' Seeding function for use with the particle filter
+##' 
+##' @title Compare model to ICU data
+##' 
+##' @param model An \code{odin_model} object
```
johnlees, line 5: A bit torn whether `odin_model` (used in models.R and less ambiguous) or just `model` (used in both sampler.R and compare.R) is best. Up to you

R/models.R
```diff
@@ -14,6 +14,7 @@ basic_model <- function(progression_groups = list(E = 2, asympt = 1, mild = 1, I
   model_class <- "sircovid_basic"
   odin_model <- load_odin_model("basic")
   generate_beta_func <- generate_beta
+  seeding_model <- seeding_function
```
johnlees, line 17: I think you need to take `seeding_pars` as an input to the models and bind that argument, as with `compare_model`

R/sircovid.R
```diff
@@ -152,10 +152,13 @@ run_particle_filter <- function(data,
   
   #set up compare for observation likelihood
   compare_func <- sircovid_model$compare_model(model = model_func, pars_obs = obs_params, data = data)
+  
+  seeding_func <- sircovid_model$seeding_model(model = model, data = data, pars_seeding = NULL)
```
johnlees, line 156: I think it makes sense for `pars_seeding` to be set in the initial model call, unless this is likely to be changed between different calls to the same model

tests/testthat/test-sampler.R
```diff
@@ -46,19 +46,21 @@ test_that("sampler runs without error", {
 
   mod <- sircovid_model$odin_model(user = pars_model)
   compare <- sircovid_model$compare_model(mod, pars_obs, d)
+  
+  seeding_func <-sircovid_model$seeding_model(mod,d)
```
johnlees, line 50: Can we add a test with `pars_seeding` set

R/seeding.R
```diff
@@ -0,0 +1,48 @@
...
+##'   \code{particle_filter_data}
+##'
+##' @param pars_seeding Parameters for seeding
+##' 
+##' @importFrom stats rpois
+##' 
+##' @export
+seeding_function <- function(model,
+                             data,
+                             pars_seeding = NULL
```
johnlees, line 17: It would be useful to demonstrate the format of `pars_seeding` somewhere. We want the default to be `NULL` so perhaps in the docstring of this function is best

R/sampler.R
edknock: I think this depends on whether the seeding_func is set up correctly. I've done a newer version of seeding.R which uses a truncated poisson function (thus avoiding a potentially long while loop) and makes clearer that the seeding times must be before the first data time point

R/seeding.R
edknock: Is this better? We use these four parameters from data https://github.com/mrc-ide/sircovid/blob/variable_seeding/R/seeding.R#L20-L23

R/sampler.R
edknock: Done

R/sampler.R
edknock: Done

R/sampler.R
edknock: Done

R/sampler.R
edknock: Done

tests/testthat/test-sampler.R
edknock: Done

R/sircovid.R
edknock: Have put pars_seeding as an argument to run_particle_filter now. Don't think we need it as an argument in beta_date_particle_filter because I think fixed seeding is inherent to that (so just need to input pars_seeding = NULL in run_particle_filter).

R/seeding.R
edknock: Done

R/models.R
johnlees:
```
basic_model <- function(progression_groups = list(E = 2, asympt = 1, mild = 1, ILI = 1, hosp = 2, ICU = 2, rec = 2),
                        gammas = list(E = 1/(4.59/2), asympt = 1/2.09, mild = 1/2.09, ILI = 1/4, hosp = 2, ICU = 2/5, rec = 2/5),
                        seeding_pars = NULL) {
  model_class <- "sircovid_basic"
  odin_model <- load_odin_model("basic")
  generate_beta_func <- generate_beta
  compare_model <- function(model, pars_obs, data) {compare_output(model, pars_obs, data, type=model_class)}
seeding_model <- function(model, data) {seeding_function(model, data, pars_seeding)}
```

edknock: I think it's better for future code (e.g. SMC^2) if we dont bind `pars_seeding`

R/seeding.R
edknock: Gone for `odin_model`

PR #74, opened 2020-04-28
Pmcmc sample both start and end date

lwhittles:
Three parameter sampler up and running, with updated tests.

I have had to stop using beta_date_particle filter in the calc_log_likelihood function, and would welcome suggestions on making this part of the code more elegant / better integrated with the rest of the package.

I have also introduced 'pars_to_sample' as an argument, which allows easy switching between the two and three parameter versions of the pmcmc and should be simple to generalise.


R/pmcmc.R
```diff
@@ -117,34 +127,38 @@ pmcmc <- function(data,
                   output_proposals = FALSE) {
   
   # test pars_init input
+  
+  par_names <- c('beta_start', 'beta_end', 'start_date')
+  par_names <- par_names[pars_to_sample[par_names]]
```
johnlees, line 132: `%in%` is more readable

R/pmcmc.R
```diff
@@ -413,17 +432,35 @@ calc_loglikelihood <- function(pars, data, sircovid_model, model_params,
...
   
+  
+  if('beta_end' %in% names(pars)) {
+    
+    new_beta <- sircovid_model$generate_beta_func(beta_start = pars[['beta_start']], 
+                                                  start_date = start_date, 
+                                                  beta_end = pars[['beta_end']])
+  } else {
+    
+    new_beta <- sircovid_model$generate_beta_func(beta_start = pars['beta_start'],
```
johnlees, line 444: Double bracket

johnlees:
I think maybe if we first make a decision about what to do with the `beta_date_particle_filter` function (leave it for the grid alone, or use it in both by adding end date) and merge this PR.

Then it's probably time to go back to #9 and think about how to infer any member of `model_params`. Maybe a class which has a range, initial value, proposal variance (and somehow covariances). beta and date are in a sense 'special' because they need functions to transform them from the parameter to use in the odin model, so we will likely keep the existing code there

tests/testthat/test-pmcmc.R
```diff
@@ -5,157 +5,189 @@ context("pmcmc")
...
     n_mcmc = n_mcmc,
     sircovid_model = sircovid_model,
     pars_obs = list(phi_ICU = 0.95,
                     k_ICU = 2,
                     phi_death = 926 / 1019,
                     k_death = 2,
-                    exp_noise = 1e6)
+                    exp_noise = 1e6),
+    pars_to_sample = c(
+      'beta_start' = TRUE,
```
johnlees, line 31: Is there a test where either `beta_start` or `start_date` are set to `FALSE` here? I didn't follow how they would get turned off

R/pmcmc.R
johnlees: Replaced in 24090f9e136c407ad4d661c59772c1d17acc56ea

PR #80, opened 2020-05-02
Add plot method for pmcmc_list (multiple chains)

lwhittles:
Small PR to add plot method for pmcmc_list objects


R/pmcmc.R
```diff
@@ -696,3 +699,165 @@ plot.pmcmc <- function(x, ...) {
...
+  n_chains <- length(chains)
+  cols_trace <- viridis::viridis(n_chains)[c(4, 1, 3, 2)]
+  
+  
+  # compile master chain and order by log posterior for plotting
+  master_chain <- do.call(what = rbind, 
+                          args = lapply(X = chains, 
+                                        FUN = function(z) {
+                                          z$results[-seq_len(burn_in), ]
+                                        }
```
johnlees, line 723: Put all these closing brackets together, or indent the last one in a clearer place

R/pmcmc.R
```diff
@@ -696,3 +699,165 @@ plot.pmcmc <- function(x, ...) {
...
+  par_names <- names(x$inputs$pars$pars_init)
+  n_pars <- length(par_names)
+  
+  chains <- x$chains
+  n_chains <- length(chains)
+  cols_trace <- viridis::viridis(n_chains)[c(4, 1, 3, 2)]
+  
+  
+  # compile master chain and order by log posterior for plotting
+  master_chain <- do.call(what = rbind,
```
johnlees, line 719: We use this function a couple of times (possibly in ncov outputs too). I think it would be useful to make into a method for `mcmc_list` and export it

R/pmcmc.R
```diff
@@ -696,3 +699,165 @@ plot.pmcmc <- function(x, ...) {
...
   
 }
+
+##' @export
+##' @importFrom viridis cividis
+##' @importFrom graphics hist par plot.new text lines legend
+##' 
+plot.pmcmc_list <- function(x, burn_in = 1, ...) {
+  
+  summ <- summary(x, burn_in = burn_in)
```
johnlees, line 709: Do you think burn in is better as a number of samples or as a proportion of the total? I did it as the proportion in ncov outputs (unless that was changed over the weekend), forgetting that you'd been using number here. It would probably be best for us to choose one way and stick with it – should it be number of samples?

johnlees:
We should also remember to replace the code in the orderly task with this new function before we next run it

R/pmcmc.R
lwhittles: I agree, it's been bugging me for a little while, thanks for the prompt

R/pmcmc.R
lwhittles: Hmm there are pros and cons to both... Shall we go with a fixed number of samples, because the part of the chain we want to discard does not vary with chain length? This will annoyingly necessitate checks on the respective lengths but I'll add those as I go.

johnlees: Sounds good – I think having the function you added above helps too, as everything should go through there now

PR #83, opened 2020-05-11
Generalise pmcmc dev

lwhittles:
Generalised pmcmc sampler now fitting to an extra six model parameters pertaining to flows through the care system

- quick bug fix in the forecasting function, forecast_days was hardcoded to zero within calc_loglikelihood
- better naming of parameters cov_mat -> proposal_kernel and default removed as getting unwieldy


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: sircovid
 Title: SIR Model for COVID-19
-Version: 0.2.27
+Version: 0.2.28
```
johnlees, line 3: Needs to be 0.2.29

PR #90, opened 2020-05-13
fix bug in update_beta

lwhittles:
bump version


edknock:
Just going to add tests for update_beta quickly

PR #111, opened 2020-05-19
Fix plotting issue

lwhittles:
small fix to plot.pmcmc_list() method to prevent warnings when n_par !=3.


johnlees:
@lwhittles let me know when this is ready for another look – the only comment on the changes might be to also export the reverse date function `sircovid_data_as_Date`

PR #112, opened 2020-05-19
fix pmcmc date bug

lwhittles:
remove date offsetting and deal with dates as 'days since 2019-12-31' in Markov chain


johnlees:
Do you think we should just remove the offset transformation in the MCMC entirely?

R/pmcmc.R
```diff
@@ -115,12 +115,11 @@ pmcmc <- function(data,
                                           0,
                                           0,
                                           0,
-                                          0,
                                           0),
                                     max=c(1,
                                           1,
                                           1,
-                                          1e6,
+                                          lubridate::yday("2020-03-15"),
```
johnlees, line 122: Use `sircovid_date()` instead

R/pmcmc.R
```diff
@@ -640,9 +637,8 @@ summary.pmcmc <- function(object, ...) {
   
   sds <- round(apply(traces, 2, sd), 3)
   # convert start_date back into dates
-  summ$start_date <- offset_to_start_date(data_start_date, summ$start_date)
-  summ$start_date <- as.Date(summ$start_date, origin="2019-12-31")
-  summ[c('2.5%', '97.5%', 'min', 'max'), 'start_date'] <- summ[c('97.5%', '2.5%', 'max', 'min'), 'start_date']
```
johnlees, line 645: Did you want to remove this line?

R/parameters.R
```diff
@@ -376,6 +376,17 @@ sircovid_date <- function(date) {
   days_into_2020
 }
 
+# Converts dates from from sircovid_Date as used in the MCMC to a Date
+# Automatically converts type
+sircovid_date_to_date <- function(sircovid_date)
```
johnlees, line 381: This should have a more informative name, perhaps `sircovid_date_as_Date()` to make the class of the return type clearer

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# sircovid 0.2.35
+
+- fix bug in pMCMC
```
johnlees, line 3: 'Remove date offsets from pMCMC'

R/parameters.R
johnlees: Also there are a two places in the plots where this same transform is used, I'd suggest switching these out to this function too:
- https://github.com/mrc-ide/sircovid/blob/master/R/sampler.R#L210
- https://github.com/mrc-ide/sircovid/blob/master/R/pmcmc.R#L644

R/pmcmc.R
lwhittles: Yes, when we used the offsets the CrI on start_date would appear the wrong way round when we converted back to dates so this was just housekeeping to deal with that. Now that we use sircovid_dates the 2.5% and 97.5% CrIs no longer need to be switched

PR #120, opened 2020-06-02
Sero flow from end of E

edknock:
The sero flow has switched from starting at the end of the I compartments to at the end of the E compartment.


R/parameters.R
```diff
@@ -222,6 +222,10 @@ generate_parameters <- function(
   }
   parameter_list$I0_asympt[seed_idx,1,parameter_list$trans_classes] <- infection_seeding$values
   parameter_list$S0[seed_idx] <- parameter_list$S0[seed_idx] - infection_seeding$values
+  if ("sircovid_serology" %in% class(sircovid_model)){
```
MJomaba, line 225: so you had the I0 in the R0_pre? on top of what the user has assigned to the R_pre?

edknock: Hopefully it is clearer now, everyone in I0_asympt goes into R0_pre (which was otherwise all 0s). This just happens in generate_parameters so happens only once. We can't do this in the odin code as we only want it to happen once at the start of the epidemic, and not every time we run the odin code (i.e. every 4 time steps).

PR #125, opened 2020-06-26
Remove start_date as required fitted parameter in PMCMC

edknock:
Adding a scaling parameter psi for importation into serology2, will make sure PMCMC runs with it (instead of start_date)


richfitz:
This *definitely* can't be merged now :joy:

PR #129, opened 2020-07-15
i113: remove interpolation functions

### LINKED ISSUE #113: Replace beta interpolation with direct calculation
We're already just computing this directly basically so it would be a look up with respect to 't'. Might be worth implementing and benchmarking as it would likely be faster.

richfitz:
This PR is half the work needed to port to dust - it removes the use of `interpolate()` with an array of values that are looked up as the model runs. This might be marginally faster, but practically it is needed because we won't want to use the interpolation functions on the GPU. The interpolate functions do capture the intent here more nicely.

Working through this, it would be nice to sit down and refactor the interface in light of where the model has grown to - this might be something to think about during the hackathon as it'll make it much easier to push this model around.

Note that for now we're using `cinterpolate::interpolation_function` to generate the vector of values but later we might replace that with R's `approx`

Fixes #113


edknock:
Added some tests to check piecewise linear beta works as we are expecting (in similar way to the standard generate_beta).

PR #131, opened 2020-07-21
i130: no output

richfitz:
This PR removes all use of output() as required by dust.

It's a bit of a fiddle because it obviously affects all reference examples, and because of a long-standing off-by-one issue with output and discrete time model.

There are several removed output variables:

* N_tot and N_tot2 - mostly used in debugging. These I've converted into ordinary variables and spat them out at the same time. They're "off" by one timestep from before but as they're meant to be constant that does not impact anything. Note that the *first* value here is zero because it's a fiddle to get the initial conditions correct and not interesting to do so.
* time - this one is easy; we just add `dt` to it to get the time at the end of the step
* beta - no longer output, and not really needed as we already have it from the input. However, I've added this as `beta_out` for now
* The values n_EI_asympt, n_EI_mild, n_EI_ILI, n_ILI_to_hosp, and n_hosp_to_ICU look to have been used only in debugging and so I've simply removed them.

One unexpected impact of this is that the model will run one fewer step per reporting interval. Through the period with data that means we'll do 4 steps rather than 5, which will actually be a significant speed bump. However, this *really* breaks the reference cases as a model without output draws a different number of random numbers than one with.

See the tree 382e76d for a version of this with a forced output step - moving to 78e208f breaks the reference examples, even though nothing of interest is changed (just because no output step is taken)


edknock:
I've switched the initial value of N_tot/N_tot2 from 0 to be user input. Thus be inputting as what we expect it to be it offers an extra check (in the tests I have switched back to checking all N_tot values are the same, and not just the 2nd value onwards). In the compare it was pulling the first value of N_tot, which with the initialisation as 0 was incorrect. This will be correct now.

edknock:
The error re N_tot in compare (when N_tot was initialised as 0) was not picked up in tests because it is only used for serology data and all the serology data were NA in the reference data. I have therefore added some serology data. I created new reference pmcmc output using the current master version of sircovid, and then confirmed that results are the same on this PR branch when an 'output' command is in the odin code. Having removed the 'output' command, I created new pmcmc reference files.

PR #136, opened 2020-09-29
Explicitly incorporate sensitivity in the serology calculation

edknock:
This changes the `sero_prob_pos` calculation so that sensitivity is now included explicitly, rather than including it implicitly within `p_seroconversion`.


edknock:
Was dealt with in PR #140

PR #139, opened 2020-10-19
Vaccination

annecori:
Added a vaccine reducing the risk of acquisition:

- one additional dimension to S for vaccinated/non vaccinated/potentially waning immunity from vaccine
- one extra parameter rel_susc describing the relative susceptibility of each vaccination group (should be 1 for the unvaccinated)
- currently no flow between the different vaccination groups (coming soon...!) so you have to be vaccinate/unvaccinated at the start and it stays that way
- currently no differential vaccine efficacy by age
- currently no effect of vaccine on risk of severe disease
- currently no effect of vaccine on risk of onward transmission

At this stage I would welcome 
- feedback on the way I've started with these very simple modifications
- feedback on whether we want to put this into a separate vaccination model or keep it as part of carehomes

Not urgent but would welcome comments!

Thanks

Anne


R/carehomes.R
```diff
@@ -46,6 +46,10 @@ NULL
 ##' @param prop_noncovid_sympt Proportion of population who do not have
 ##'   covid but have covid-like symptoms
 ##'
+##' @param rel_susc A vector of values representing the relative susceptibility
```
richfitz, line 49: I wonder if we should expand this name a little? `rel_susceptibility` perhaps? Or something related to vaccination?

R/carehomes.R
```diff
@@ -68,10 +72,15 @@ carehomes_parameters <- function(start_date, region,
                                  react_specificity = 0.99,
                                  react_sensitivity = 0.99,
                                  prop_noncovid_sympt = 0.01,
+                                 rel_susc = 1,
                                  exp_noise = 1e6) {
   ret <- sircovid_parameters_shared(start_date, region,
                                     beta_date, beta_value)
 
+  ## TO DO: we should add some checks for pillar2_specificity,
```
richfitz, line 80: Can you either add these or open a ticket that explains the problem more fully and reference it here?

R/carehomes.R
```diff
@@ -428,14 +439,15 @@ carehomes_initial <- function(info, n_particles, pars) {
   index_N_tot3 <- index[["N_tot3"]][[1L]]
 
   index_S <- index[["S"]]
+  index_S_no_vacc <- index_S[seq_len(length(pars$N_tot))]
```
richfitz, line 442: This implies that you have organised your S like:

```
S[i, j]
```

where `i` is the age group and `j` is the vaccination level, right?

R/carehomes.R
```diff
@@ -448,6 +460,20 @@ carehomes_initial <- function(info, n_particles, pars) {
 }
 
 
+##' Carehomes vaccination parameters.
+##'
+##' @title Carehomes vaccination parameters
+##'
+##' @return A list of parameter values
+##'
+##' @export
```
richfitz, line 469: I don't think that this needs exporting, nor documenting with roxygen

R/carehomes.R
```diff
@@ -448,6 +460,20 @@ carehomes_initial <- function(info, n_particles, pars) {
...
+carehomes_parameters_vaccination <- function(rel_susc = 1) {
+  list(N_vacc_classes = length(rel_susc),
```
richfitz, line 471: Do you make any assumptions about the patterns of `rel_susc`? Must it be monotonic if it has a length of greater than 1 for example?

inst/odin/carehomes.R
```diff
@@ -1,22 +1,23 @@
...
+## k for the infectivity group (for I) or vacc. group (for S)
 
-## Number of age classes & number of transmissibility classes
+## Number of classes (age transmissibility & vaccination)
 
 ## TODO: this should be renamed as it includes the CHW and CHR groups,
 ## so it's N_age plus 2 now! N_group is ok but more vague than ideal.
 N_age <- user()
 trans_classes <- user(1)
+N_vacc_classes <- user(1)
```
richfitz, line 12: There's a dependency here with rel_susc, so this should be:

```
N_vacc_classes <- length(rel_susc)
```

and then remove the parameter from the R code

inst/odin/carehomes.R
```diff
@@ -1,22 +1,23 @@
...
 
 ## Definition of the time-step and output as "time"
 dt <- user()
 initial(time) <- 0
 update(time) <- (step + 1) * dt
 
 ## Core equations for transitions between compartments:
-update(S[]) <- S[i] - n_SE[i]
+update(S[, ]) <- S[i, j] - n_SE[i, j] # age / vaccination status
```
richfitz, line 20: can you use a comma in the comment so it does not look like division?

inst/odin/carehomes.R
```diff
@@ -482,6 +483,10 @@ initial(cum_admit_by_age[]) <- 0
 
 ## User defined parameters - default in parentheses:
 
+## Parameters of the S classes
+rel_susc[] <- user()
+dim(rel_susc) <- N_vacc_classes
```
richfitz, line 488: Ah I see why you had the above. The way to do this nicely with odin (all 3 lines) is:

```r
rel_susc[] <- user()
dim(rel_susc) <- user() # this says - take the length as whatever was provided by the user
N_vacc_classes <- length(rel_susc)
```

richfitz (changes requested):
This looks good and the changes are small enough that folding them into the main model seems reasonable. I like the tests. 

The failing build is caused by warnings from R CMD check - if you remove the new internal function from the exported set this one will disappear:

```
* checking for missing documentation entries ... WARNING

Undocumented code objects:

  'carehomes_parameters_vaccination'
```

(most likely you have an uncommited file in man/ at the moment)

R/carehomes.R
annecori: yes that's right

R/carehomes.R
annecori: Typically we will have 
rel_susc <- c(1, vacc, waning_1, waning_2, ...) 
where vacc < waning_1 < waning_2 < ... < 1
I will write this in a function which will check the format is that expected and give a warning if not I think, if that seems sensible?

richfitz: yes, sounds good

richfitz: Ah, actually no - don't give a warning, please throw an error

richfitz (changes requested):
One more thing - can you please increase the version number; it must be increased at every PR as the merge to master triggers a binary build

R/carehomes.R
```diff
@@ -428,14 +436,15 @@ carehomes_initial <- function(info, n_particles, pars) {
   index_N_tot3 <- index[["N_tot3"]][[1L]]
 
   index_S <- index[["S"]]
```
richfitz, line 438: This is going to require some changes in `sircovid::carehomes_Rt_trajectories`, Anne suggests a weighted sum (relative to susceptibility/efficacy) over the S categories to give the same size S vector as before. Attention: @edknock

R/carehomes.R
annecori: @richfitz has now recoded these checks

R/carehomes.R
annecori: @richfitz and @edknock I have now implemented changed in `sircovid::carehomes_Rt_trajectories` to allow for Rt calculation with the additional layer of stratification for S (now S has 2 dimensions, for age and vaccination status). 
I think this works but I've run into issues with tests which were using pre-recorded data for which the `rel_susceptibility` parameter was absent. I've artificially added it back with its default value of `1`, but that's pretty ugly. @richfitz if you can tell me how to better address this it would be great.
And @edknock if you could please check you agree with the changes I've made the the Rt calculation that would be much appreciated. 
Thanks!

annecori: @richfitz see for example [this:](https://github.com/mrc-ide/sircovid/blob/5921e5f091746c8572f7165d7641df1967bca0b4/tests/testthat/test-carehomes-rt.R#L10)

annecori: and @edknock please see [here](https://github.com/mrc-ide/sircovid/blob/5921e5f091746c8572f7165d7641df1967bca0b4/R/carehomes_rt.R#L37)

richfitz (approved):
Looks good to me so long as it is ok for the Rt calculation

edknock:
Rt calculation looks good. We could add a deterministic test that if you have the same susceptible population size by age and we calculate Rt when (a) everyone is unvaccinated and (b) some are vaccinated with lower susceptiblity, that we get a lower Rt for (b).

annecori:
@edknock and @richfitz I have addressed all your questions/comments. 
Please check you are happy with my latest test addition [here](https://github.com/mrc-ide/sircovid/blob/d70dc0da0ad64e51776de53ffbd2f91900bce87f/tests/testthat/test-carehomes-vaccination.R#L79).
And if you are happy with this can one of you please merge to master (pending check success)?

richfitz (approved):
LGTM, nice idea for a test @edknock

PR #140, opened 2020-10-21
Introduce test_prob_pos, tweak sero calculation and wire in pillar 2 over 25

edknock:
Since we now have several test data streams, all requiring calculation of the probability of a positive test, I figured it was better to create a function that does this, which can then be tested.

This function also now adds exponential noise to the negatives, so that now we also ensure the probability is less than 1 as well as more than 0.

As part of this I have also tweaked the sero probability calculation. Now we explicitly consider sensitivity (rather than hiding it implicitly in p_seroconversion), and as a result we now apply specificity to R_neg and R_pre individuals.

I have also wired in fitting pillar 2 for only the over 25s as an option, and set it so that running the particle filter with more than one pillar 2 datastream throws an error.


R/carehomes.R
```diff
@@ -254,26 +259,44 @@ carehomes_compare <- function(state, prev_state, observed, pars) {
...
-  react_pos <- model_react_pos +
-    rexp(length(model_react_pos), pars$observation$exp_noise)
-  react_neg <- sum(pars$N_tot[2:18]) - model_react_pos
-  model_react_prob_pos <- (pars$react_sensitivity * react_pos +
-                             (1 - pars$react_specificity) * react_neg) /
-    (react_pos + react_neg)
+  ## calculate test positive probabilities for the various test data streams
+  ## Pillar 2
+  pillar2_negs <- pars$prop_noncovid_sympt * (sum(pars$N_tot)
+                                              - model_sympt_cases)
```
lwhittles, line 271: This is perhaps a little specific - we're removing model_sympt_cases, but not those in hospital or those who have died or care home residents, who are likely to be greater in number than the model_sympt_cases... what do you think?

R/support.R
```diff
@@ -83,6 +83,24 @@ dbetabinom <- function(x, size, prob, rho, log = FALSE) {
 }
 
 
+##' @importFrom stats rexp
+test_prob_pos <- function(pos, neg, sensitivity, specificity, exp_noise) {
```
lwhittles, line 87: This is so much simpler - a huge improvement

R/carehomes.R
```diff
@@ -254,26 +259,44 @@ carehomes_compare <- function(state, prev_state, observed, pars) {
...
+  model_pillar2_prob_pos <- test_prob_pos(model_sympt_cases,
+                                          pillar2_negs,
+                                          pars$pillar2_sensitivity,
+                                          pars$pillar2_specificity,
+                                          pars$observation$exp_noise)
+
+  ## Pillar 2 over 25s
```
lwhittles, line 278: Should we be including CHR here in general - would come under pillar 1 testing? We exclude them from REACT so we should check if pillar 2 should take the same approach

R/carehomes.R
```diff
@@ -325,13 +342,24 @@ carehomes_compare <- function(state, prev_state, observed, pars) {
...
                                 pars$k_pillar2_cases, exp_noise)
 
+  ll_pillar2_over25_tests <- ll_betabinom(observed$pillar2_over25_pos,
+                                          observed$pillar2_over25_tot,
+                                          model_pillar2_over25_prob_pos,
+                                          pars$rho_pillar2_tests)
+
+  ll_pillar2_over25_cases <- ll_nbinom(observed$pillar2_over25_cases,
+                                       pars$phi_pillar2_cases *
+                                         model_sympt_cases,
```
lwhittles, line 352:
```suggestion
                                         model_sympt_cases_over25,
```

R/carehomes.R
```diff
@@ -254,26 +259,44 @@ carehomes_compare <- function(state, prev_state, observed, pars) {
...
+
+  ## serology
+  model_sero_prob_pos <- test_prob_pos(model_sero_pos,
+                                       pars$N_tot_15_64 - model_sero_pos,
+                                       pars$sero_sensitivity,
+                                       pars$sero_specificity,
+                                       pars$observation$exp_noise)
 
   pars <- pars$observation
   exp_noise <- pars$exp_noise
```
lwhittles, line 302: Could maybe move this up, say to line 267 so you can use the unpacked exp_noise object throughout the calls to test_prob_pos()

This is v. minor though!

R/carehomes.R
```diff
@@ -627,11 +655,25 @@ carehomes_particle_filter_data <- function(data) {
...
+                "pillar2_over25_cases", "react_pos", "react_tot")
+
   verify_names(data, required, allow_extra = TRUE)
+
   if (any(!is.na(data$deaths) &
            (!is.na(data$deaths_comm) | !is.na(data$deaths_hosp)))) {
     stop("Deaths are not consistently split into total vs community/hospital")
   }
+
+  pillar2_streams <- sum(c(any(!is.na(data$pillar2_pos)) |
```
lwhittles, line 668: This looks correct, but it took me a while!

inst/odin/carehomes.R
```diff
@@ -880,16 +875,17 @@ initial(sero_prob_pos) <- 0
...
+update(sero_pos) <- sum(new_R_pos[4:13])
 
 initial(cum_sympt_cases) <- 0
 update(cum_sympt_cases) <- cum_sympt_cases + sum(n_EI_mild) + sum(n_EI_ILI)
 
-## For REACT we exclude the 0-4 and CHR groups
+## only over 25s (exclude groups 1 to 5)
+initial(cum_sympt_cases_over25) <- 0
+update(cum_sympt_cases_over25) <- cum_sympt_cases_over25 +
+  sum(n_EI_mild[6:19, ]) + sum(n_EI_ILI[6:19, ])
```
lwhittles, line 887: See earlier comment re: inclusion of CHR in pillar2

lwhittles (changes requested):
This is a big improvement on the previous apporach, much clearer and easier to intepret, especially wrt to the handling of sensitivity and specificity. I found one typo in the over25 likelihood but otherwise all good. We should also have a collective chat about whether to include CHR in pillar 1 - I will cross post this in Teams as it would be good to get this nailed down within this package update

lwhittles (approved):

PR #143, opened 2020-10-27
Seroreversion

edknock:
Adding seroreversion to the carehomes model. Individuals in the sero flow stay in `R_pos` for a duration of Erlang distribution, after which they move to `R_neg`.

Note: I also added the 10 initial infectives into the initial value of `react_pos`, as they should be included there initially (though this would've had no impact).


lwhittles (approved):
Looks great, is it maybe worth adding a check that if we can turn sero-reversion off and on, i.e. by setting p_seroconversion = 1, and then gamma_R_pos to Inf?

I'll approve in the meantime so you can implement if you want but it doesn't hold anything up

edknock:
> Looks great, is it maybe worth adding a check that if we can turn sero-reversion off and on, i.e. by setting p_seroconversion = 1, and then gamma_R_pos to Inf?
> 
> I'll approve in the meantime so you can implement if you want but it doesn't hold anything up

This requires gamma_R_pos = 0 (so duration in R_pos is infinite). I already had to add this in to the 'no-one does not seroconvert test' to ensure no-one enters R_neg 0. I have edited the name of that test to clarify this.

PR #144, opened 2020-11-02
Waning immunity

annecori:
Adds a flow from R to S (in the unvaccinated class). 
The number of individuals selected through binomial draw to move from R to S is then also removed from R_neg and PCR_neg (as discussed with @MJomaba following the assumption that individuals loosing immunity are among those who lost antibodies - and test PCR negative)


edknock:
Looks good. Deals well with how waning immunity affects the sero/PCR flows.

I think we have waning_rate = 0 in all the test-carehomes-check tests, and feel it would probably be better to test with a non-zero value, to make sure waning immunity doesn't cause anything unexpected. However I expect some specific tests may still require waning_rate = 0.

R/carehomes.R
```diff
@@ -53,6 +53,11 @@ NULL
 ##'   value should be 1 (for the non-vaccinated group) and subsequent values be
 ##'   between 0 and 1.
 ##'
+##' @param waning_rate A single value or a vector of values representing the
```
richfitz, line 56: Are we ok with it only being immunity that wanes? (i.e., is this descriptive enough?)

R/carehomes.R
```diff
@@ -77,6 +82,7 @@ carehomes_parameters <- function(start_date, region,
                                  react_sensitivity = 0.99,
                                  prop_noncovid_sympt = 0.01,
                                  rel_susceptibility = 1,
+                                 waning_rate = 0,
```
richfitz, line 85: Just to confirm: a zero rate collapses to the current model?

R/carehomes.R
```diff
@@ -509,6 +517,15 @@ carehomes_parameters_vaccination <- function(rel_susceptibility = 1) {
 }
 
 
+carehomes_parameters_waning <- function(waning_rate = 0) {
+  waning_rate <- build_waning_rate(waning_rate)
+  list(
+    # leaving this function as may add more vaccination parameters later
```
richfitz, line 523: this comment is incorrect, but can be deleted generally as there's no apology to be made here

R/carehomes.R
```diff
@@ -712,3 +729,7 @@ carehomes_particle_filter_data <- function(data) {
 
   data
 }
+
+get_n_groups <- function() {
```
richfitz, line 733: This should be `carehomes_n_groups()` as it's specific to the carehomes model

R/utils.R
```diff
@@ -93,3 +93,22 @@ check_rel_susceptibility <- function(rel_susceptibility) {
       "increasing"))
   }
 }
+
+build_waning_rate <- function(waning_rate) {
+  if (length(waning_rate) == 0) {
+    stop("At least one value required for 'rel_susceptibility'")
+  }
+  N_age <- get_n_groups()
```
richfitz, line 101: the N_age / n_groups split here is jarring

R/utils.R
```diff
@@ -93,3 +93,22 @@ check_rel_susceptibility <- function(rel_susceptibility) {
...
+  if (any(waning_rate < 0)) {
+    stop("'waning_rate' must have only non-negative values")
+  }
+  if (length(waning_rate) > 1) {
+    if (length(waning_rate) != N_age) {
+      stop("'waning_rate' should have as many elements as age groups")
+    }
+    vect_waning_rate <- waning_rate
+  } else { # create vector by repeating waning_rate for each age group
```
richfitz, line 110: This comment can be dropped unless you feel strongly. Alternatively, write this block as:

```r
if (length(waning_rate) == 1) {
  waning_rate <- rep(waning_rate, N_age)
} else if (length(waning_rate) != N_age) {
  stop("'waning_rate' should have as many elements as age groups, or be a scalar")
}
waning rate
```

(You do not need the `each`, taking the scalar branch first makes the logic clearer, and avoiding the rename means less is going on)

tests/testthat/test-carehomes-check.R
```diff
@@ -48,6 +48,38 @@ test_that("everyone is infected when beta is large", {
 })
```
richfitz, line 50: Is it worth adding a test that if waning rate is zero then noone ever goes back? You might need to turn off some other processes to make this really obvious, and it might not be worthwhile. We do have a test of this already to a degree in that the reference example tests are passing (so you do have the same model!)

tests/testthat/test-carehomes-check.R
```diff
@@ -48,6 +48,38 @@ test_that("everyone is infected when beta is large", {
...
+  state[index_S] <- 0
+
+  mod$set_state(state)
+  mod$set_index(integer(0))
+  y <- mod$transform_variables(drop(
+    dust::dust_iterate(mod, seq(0, 400, by = 4))))
+
+  # other than in the 4th age group (where infections are seeded)
+  # after the first time step, R is empty
+  expect_true(all(y$R[-4, -1] == 0))
```
richfitz, line 76: if only one event happens in any given timestep, surely people should spent one timestep in R?

richfitz (changes requested):
Someone else should look at the biology side, but this looks simple enough from the changes to the model (so long as it fits in with the seroreversion, but I know you're already discussing that)

edknock:
I've set waning_rate to be nonzero here to ensure the N_tot tests still hold with waning immunity https://github.com/mrc-ide/sircovid/blob/waning_immunity/tests/testthat/test-carehomes-check.R#L6

tests/testthat/test-carehomes-check.R
```diff
@@ -48,6 +50,56 @@ test_that("everyone is infected when beta is large", {
 })
 
 
+test_that("noone stays in R, R_neg or PCR_neg if waning rate is very large", {
+  p <- carehomes_parameters(0, "england", beta_value = 0,
+                            waning_rate = 1e9)
```
edknock, line 55: Could set `waning_rate = Inf` here, as typically all tests for rates (gammas) are at 0 or Inf. The test where beta is very large doesn't use a value of Inf because in that specific case as it causes NA values in compartments.

R/carehomes.R
edknock: Yes, zero rate means people will stay in R and never move back to S.

R/utils.R
edknock: Might be worth finally correcting this so that `N_groups` replaces `N_age` where really we mean `N_age + 2`?

edknock:
Happy with this from a modelling POV. Have tweaked test-carehomes-check, so @annecori can you look and see what you think, feel free to tweak the `waning_rate = 0` test further if you have a better idea.

tests/testthat/test-carehomes-check.R
```diff
@@ -48,6 +50,56 @@ test_that("everyone is infected when beta is large", {
...
+  mod$set_state(state)
+  mod$set_index(integer(0))
+  y <- mod$transform_variables(drop(
+    dust::dust_iterate(mod, seq(0, 400, by = 4))))
+
+  # other than in the 4th age group (where infections are seeded)
+  # after the first time step, R is empty
+  expect_true(all(y$R[-4, -1] == 0))
+  # so is R_neg
+  expect_true(all(y$R_neg[-4, -1] == 0))
```
edknock, line 80: Should there also be a check here for `PCR_neg`?

R/carehomes.R
annecori: @edknock what do you think?

R/carehomes.R
annecori: yes

R/utils.R
annecori: suggest doing this in separate PR. At the moment have just removed "N_age" from this specific function

tests/testthat/test-carehomes-check.R
annecori: right - meant day, not timestep - there are 4 time steps per day as far as I can tell. 
Maybe @edknock you can confirm I understand correctly that
`dust::dust_iterate(mod, seq(0, 400, by = 4)`
runs the model mod for 100 days with time step 1/4 day?

tests/testthat/test-carehomes-check.R
annecori: @edknock has added something like this [here](https://github.com/mrc-ide/sircovid/blob/802a521995d03910d2fac191d40b7e32ec48a8dd/tests/testthat/test-carehomes-check.R#L84)

tests/testthat/test-carehomes-check.R
richfitz: things should still transition through R though? if it's always empty that sounds like a problem to me

R/carehomes.R
edknock: I think it's fine.

tests/testthat/test-carehomes-check.R
annecori: so I initiate the model so everyone is in R instead of S (apart from age group 4 where the seeding is done so some people are in E or I and I didn't want to bother with that). In all other age groups everyone usually starts in S, but I've moved them all to R initially, then with a large waning rate they should all leave R immediately.

annecori: I have tried to clarify in comments in the test, see [here](https://github.com/mrc-ide/sircovid/blob/ebcb2cda219dbf496258e553d898efb02243d3c8/tests/testthat/test-carehomes-check.R#L53).

annecori:
> Happy with this from a modelling POV. Have tweaked test-carehomes-check, so @annecori can you look and see what you think, feel free to tweak the `waning_rate = 0` test further if you have a better idea.

@edknock I have changed most tests in test-carehomes-check and test-carehomes-vaccination to use a non null waning rate when possible (sometimes the tests were designed specifically based on the assumption that there is no movement from R to S in which case I left the default zero waning rate). 

I had issues with the test for `R_pre parameters work as expected` which failed when waning_rate was not zero, could you please check that you understand why that's the case ? Not sure what this test does...

The other thing I haven't tested with waning immunity is all the calculations of Rt. Would you be happy if I reset the reference dataset so that all these test have a non null waning rate as well? Or should we leave as is? What do you think?

Other than this I have addressed all your comments and those from @richfitz as well.

edknock:
> I had issues with the test for `R_pre parameters work as expected` which failed when waning_rate was not zero, could you please check that you understand why that's the case ? Not sure what this test does...

This just checks the time to seroconversion parameters work correctly. It's the removal of individuals from `R_neg` that breaks this when you have a nonzero waning rate. Since this test is specifically about testing the parameters governing the `R_pre` classes, it's fine to just use a zero waning rate in this case.

edknock:
> The other thing I haven't tested with waning immunity is all the calculations of Rt. Would you be happy if I reset the reference dataset so that all these test have a non null waning rate as well? Or should we leave as is? What do you think?

The only effect on the Rt calculation is that the susceptible population can go up as well as down now, and it should have no problem dealing with this. Not sure you need to change the reference dataset, but there's no harm in doing so.

richfitz (approved):

annecori:
> > The other thing I haven't tested with waning immunity is all the calculations of Rt. Would you be happy if I reset the reference dataset so that all these test have a non null waning rate as well? Or should we leave as is? What do you think?
> 
> The only effect on the Rt calculation is that the susceptible population can go up as well as down now, and it should have no problem dealing with this. Not sure you need to change the reference dataset, but there's no harm in doing so.

agreed so I don't think we need to edit these tests

edknock:
> > > The other thing I haven't tested with waning immunity is all the calculations of Rt. Would you be happy if I reset the reference dataset so that all these test have a non null waning rate as well? Or should we leave as is? What do you think?
> > 
> > 
> > The only effect on the Rt calculation is that the susceptible population can go up as well as down now, and it should have no problem dealing with this. Not sure you need to change the reference dataset, but there's no harm in doing so.
> 
> agreed so I don't think we need to edit these tests

Yes I think if `carehomes-rt` is unchanged, we generally don't need to edit the tests for it.

PR #147, opened 2020-11-05
Rename N_age to n_groups

### LINKED ISSUE #146: Use n_groups not N_age in the carehomes model

richfitz:
As discussed, this does a mass rename of N_age in the carehomes model to n_groups. We retain N_age in the parameters set, which is nice as the Rt calculation becomes a little easier to read. Note that this required regenerating the rt.rds file as otherwise the new parameter is not present in the inputs.

Questions:
* I've gone with lowercase `n_` here - do we want to change that for N_age too? In contrast you also have N_tot which is a count across compartment which feels more like a real upper case N
* This will break the ability to compute Rt or do simulations against previously saved data; do we need any sort of compatibility layer here?

Master is failing due to the last PR not having the version number bumped so we should try and merge this in fairly smoothly.

Update: This is now a hugely noisy diff, but is just set of of renames:
* N_age => n_age_groups / n_groups (depending on use)
* trans_classes => n_trans_classes

I have added a NEWS.md file to help us keep track of things

Fixes #146


edknock:
Is it worth using n_groups/N_age here now?
https://github.com/mrc-ide/sircovid/blob/i146-n-groups/inst/odin/carehomes.R#L457-L458

edknock (commented):
Looks good to me. I think fine to go with `n_age` for consistency with `n_groups`.

NEWS.md
```diff
@@ -0,0 +1,22 @@
...
+* Implement preliminary support for seroreversion
+
+# sircovid 0.6.3
+
+* Implement preliminary support for vaccination modelling
+
+# sircovid 0.6.2
+
+* Support for probability of positive calculation directly in the package
+* Pillar 2 data is now fitted to over 25s only
```
edknock, line 17: Maybe reword this as we can still fit Pillar 2 data to all ages. Over 25s only is a new, alternative option.

NEWS.md
```diff
@@ -0,0 +1,22 @@
+# sircovid 0.6.5
+
+* Parameter `N_age` was split into `n_age` (both models) and `n_groups` (carehomes)
```
edknock, line 3: I guess n_age here should now be n_age_groups?

inst/odin/carehomes.R
```diff
@@ -454,8 +455,8 @@ I_with_diff_trans[, ] <-
 ## NOTE: "age groups" 1-17 are age groups, 18 are CHW and 19 CHR. Here we apply
 ## beta to all contacts *except* within care home contacts
 s_ij[, ] <- m[i, j] * sum(I_with_diff_trans[j, ])
-s_ij[1:17, 1:19] <- beta * s_ij[i, j]
-s_ij[18:19, 1:17] <- beta * s_ij[i, j]
+s_ij[1:n_age_groups, 1:n_groups] <- beta * s_ij[i, j]
+s_ij[(n_age_groups + 1):n_groups, 1:n_age_groups] <- beta * s_ij[i, j]
```
edknock, line 459: Perhaps the comment about this section needs editing now?

edknock (approved):
Just a few suggestions re NEWS and comments, otherwise I am happy with this

inst/odin/carehomes.R
richfitz: I thought that comment was still reasonable

NEWS.md
richfitz: thanks!

PR #142, opened 2020-10-23
Vaccination by age

annecori:
This builds on (and modifies) #139.

- Relative susceptibility due to vaccine can now be age dependant

- Vaccination and building up/waning of immunity now implemented, i.e. individuals can move between different S compartments, with parameters which can be age dependant. 

- Parameters governing movement between S compartments are not time varying at the moment

- For this to work I had to force having at least 3 vaccination classes, but by default the second and third and flows to those are empty / zero. NOTE THIS IS LIKELY TO BREAK THE ORDERLY TASKS

Would be great to discuss and if we want to make this part of master, merge asap before master diverges again.

PR #151, opened 2020-11-10
splitting I compartments to differentiate vaccination status

annecori:
**New features:**

- used the stratification by different infectivity levels to model different vaccination strata for E and I compartments instead
- Added stratification of R compartments by vaccination status as well 
- so in this version we have stratification by vaccination status for all compartments (except dead)

- `vaccination_rate` has been removed and is now included as first element of `vaccine_progression_rate`
- added a flow back from last vaccination stage to first so we can use vaccination structure to model things like isolation [currently only exists for S]. This is the last element of `vaccine_progression_rate`. This fixes issue #152 
- as a result the minimum number of vaccination classes is 1 instead of 3

- we now count for the fact that an individual in an S compartment can be infected and vaccinated at the same timestep

**Note:**

This is likely to break the orderly tasks because by default the number of vaccination classes is now back to 1, not 3.

**Not yet implemented:**

- vaccination flow for compartments other than S. At the moment only individuals in S can be vaccinated and move through vaccination strata
- time varying `vaccine_progression_rate`
- different `vaccine_progression_rate` for S, E, I, R etc (in particular no vaccination of hospitalised cases)
- new parameters for differential probability of asymptomatic and of severe given symptomatic according to vaccination strata
- new parameter for differential infectivity by vaccination status


richfitz:
@annecori as #142 is merged now, you should merge master into `vaccination_I`

```
git fetch
git merge origin/master
git push
```

which will update the diff

PR #155, opened 2020-11-11
Add death in stepdown to carehomes model

edknock:
This splits the `I_triage_R -> I_ICU_R -> R_stepdown -> R` pathway into two:

1. `I_triage_S -> I_ICU_S_R -> R_stepdown_R -> R`
2. `I_triage_S -> I_ICU_S_D -> R_stepdown_D -> D_hosp`

With subscript S here now for stepdown. Open to renaming here (`R_stepdown_D` is perhaps confusing).

I added to test-carehomes-check to test for the new parameters. Have checked that when parameters are set to replicate the previous version of the model we get very similar results (but not the same - confident this is due to reordering of the draws). I also checked the new Rt calculation gives the same values when the parameters replicate the previous version.


lwhittles:
This is probably very annoying, but, does it make sense to split _after_ triage rather than before, so we only have a single triage compartment (as per the model diagram) rather than two?

triage -> ICU_S_R -> R
           -> ICU_S_D -> D_hosp
           -> ICU_D     -> D_hosp

could do this later in a separate PR if you'd rather get this moving?

edknock:
> This is probably very annoying, but, does it make sense to split _after_ triage rather than before, so we only have a single triage compartment (as per the model diagram) rather than two?
> 
> triage -> ICU_S_R -> R
> -> ICU_S_D -> D_hosp
> -> ICU_D -> D_hosp
> 
> could do this later in a separate PR if you'd rather get this moving?

Have done this now

richfitz (commented):
I'm fine with this from a code point of view, looks like a ton of work.

@annecori, I presume this is more integration pain for you?

MJomaba (approved):
Looking good!

Issue #152, opened 2020-11-10
allow flow back to unvaccinated from last vaccinated stage


annecori:
for all S, E, I, R compartments

richfitz:
I think this is done now @annecori ?

annecori:
yes!

PR #156, opened 2020-11-14
Vaccination (1/3) - new compartments and flows

annecori:
Screenshot 2020-11-16 at 08 48 59

**New features:**

The diagram above shows how the structure and flows have changed on this branch compared to master. New features include:

- used the stratification by different infectivity levels to model different vaccination strata for E and I_asympt compartments instead

- Added stratification of R compartments by vaccination status as well

- so in this version we have stratification by vaccination status for all compartments except: dead, I_mild, I_ILI, I_comm_D, and all the hospitalised compartments. Rationale for this is that we won't vaccinated individuals who have symptoms or are in hospital.

- vaccination_rate has been removed and is now included as first element of vaccine_progression_rate

- added a flow back from last vaccination stage to first so we can use vaccination structure to model things like isolation. This is the last element of vaccine_progression_rate. This fixes issue #152

- as a result the minimum number of vaccination classes is 1 instead of 3

- we now count for the fact that an individual in an S compartment can be infected and vaccinated at the same timestep. Same for individuals in E, I_asympt or R.

**Note:**

This is likely to break the orderly tasks because by default the number of vaccination classes is now back to 1, not 3.

**Not yet implemented:**

time varying vaccine_progression_rate
different vaccine_progression_rate for S, E, I, R 
new parameters for differential probability of asymptomatic and of severe given symptomatic according to vaccination strata
new parameter for differential infectivity by vaccination status


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: sircovid
 Title: SIR Model for COVID-19
-Version: 0.6.10
+Version: 0.6.11
```
richfitz, line 3: Can you add an update to the NEWS too please?

tests/testthat/test-carehomes-vaccination.R
```diff
@@ -31,15 +31,14 @@ test_that("there are no infections if everyone is vaccinated with a vaccine
   expect_true(all(s[, 2, ] == s[, 2, 1]))
 })
 
-test_that("Everyone moves to vaccinated and stays there if everyone quickly
-          gets vaccinated with a vaccine preventing 100% of acquisition
+test_that("Every susceptible moves to vaccinated and stays there if everyone
```
richfitz, line 34: Can you shrink these really long explanations so that they fit on one line, so that the test indentation is not so weird? I know this is an annoying thing, but it's the stuff that accumulates over time (You will need to tell Rstudio to re-indent after doing this, that will make your lines much easier to deal with as you'll save ~10 characters)

If you need to explain the test more you can use a comment - this small string is the thing printed when the test fails, so should not be a paragraph in any case (affects a few others I believe)

richfitz (changes requested):
This looks fine to me - I'll leave it to @edknock to make sure that the epi is ok :)

A couple of things:

* Can we tidy up the test comments (see below)
* Perhaps this should become 0.7.0 as it's a fairly major new feature
* For the genre of tests "Set rates huge and confirm flows" we should be able to rig up tests where you configure the population and move things through exactly one step. With this you can be much more targeted with the effects. Let me know if that would be useful to explore and I can help

edknock:
I am happy with this. Are there changes left to be made for this to be merged?

richfitz (approved):

PR #157, opened 2020-11-16
Vaccination parameters

annecori:
added a parameter `rel_p_sympt` which depends on age (first dimension) and vaccination class (second dimension) and acts as a multiplier on the probability of being symptomatic

Builds on #156, so see diff: 
https://github.com/mrc-ide/sircovid/compare/vaccination_flows_restart...vaccination_parameters


R/vaccination.R
```diff
@@ -1,79 +1,59 @@
-build_rel_susceptibility <- function(rel_susceptibility) {
+build_rel_param <- function(rel_param, name_param = "rel_param") {
```
richfitz, line 1: my strong suggestion is to avoid default args on internal functions like this, as it leads to surprising bugs when you forget them

R/vaccination.R
```diff
@@ -1,79 +1,59 @@
...
   }
-  mat_rel_susceptibility
+  mat_rel_param
 }
 
 
-check_rel_susceptibility <- function(rel_susceptibility) {
-  if (length(rel_susceptibility) == 0) {
-    stop("At least one value required for 'rel_susceptibility'")
+check_rel_param <- function(rel_param, name_param = "rel_param") {
```
richfitz, line 19: Again, I'd avoid a default here

richfitz (changes requested):
This looks fine on the code side, except the default argument. If this is outstanding before #156 is merged we might merge this into that branch and then do one merge into master.

PR #166, opened 2020-11-25
Correct Rt calculation for p_rel_sympt and p_rel_hosp_if_sympt

edknock:
The Rt calculation had taken into account `rel_susceptibility` for vaccinated individuals but not yet `p_rel_sympt` or `p_rel_hosp_if_sympt`. This corrects this.

Also introduces a function into utils, `matrix_pow`, which is used in the Rt calculation. Was named `matrix_exp` (for exponent) at first, but after discussion with @annecori, decided to not use that at first as might have been confused for a function calculating a matrix exponential (which is a different thing altogether).

We may want to add some tests for Rt that check e.g. that having vaccine classes with lower `p_rel_sympt`/`p_rel_hosp_if_sympt` results in lower Rt.


annecori (approved):

richfitz (approved):

annecori (commented):
new tests look good!

PR #174, opened 2020-12-16
Doc typo fix

seabbs:
Great work on this package. 

On reading through some of documentation spotted and fixed a couple of typos. 

Sam

P.S I realise this is WIP but would suggest a slight reword of the `carehomes_forecast` documentation
P.P.S It would be really nice to have a Github actions built docker container/binder link to a working install 😄


richfitz:
Thanks @seabbs !

The build failure here is because I need to regenerate the interface to match the changes introduced in dust. We also need to regenerate the roxygen docs. I've just pushed up some fixes to your branch for that.

I have a dockerfile somewhere but never organised building it here, as our deployment is a little more involved. I've opened #175 to fix this.

seabbs:
No problem! 

Make sense. I always panic and need a Dockerfile when install instructions are longer than 3 lines 😆.

richfitz (approved):

PR #178, opened 2020-12-23
Multistrain

annecori:
Add ability to cope with multiple strains. Features are, in short:
-	2 strains only at the moment; easy to increase number of strains but need adapting the code for seeding the additional strains
-	No coinfection
-	Seeding of the new strain: done by moving a fixed number of unvaccinated indivuduals aged 15-19 from S to E[new_strain] each day over a pre-defined seeding period
-	New strain is characterised by relative transmissibility compared to strain 1, same across all ages
-	Severity and clinical progression all same as strain 1 
-	We track the strain in all compartments (except S). 

**STILL TO DO:**

- [x] check that the indexing is still working as we've added a layer of stratification
- [ ] ~check the NGM and R calculations - this was coded assuming that the R0 is the max of all strain specific R0. @edknock  please check you are happy with that.~ (moved to #184)
- [x] consider reordering of indexes to something more logical - especially in anticipation of having to add one more layer of stratification for the multiple vaccines
- [x] identify memory error


richfitz:
That failure on ubuntu/R-devel is legit: valgrind is reporting some issues. I will investigate but it's likely not that bad. I've checked with 0.8.0 and there are no memory issues there, so this does look newly introduced (and also not an issue for any current runs)

inst/odin/carehomes.R
```diff
@@ -173,545 +174,595 @@ prob_admit_conf[] <- p_admit_conf * psi_admit_conf[i]
 #### flow out of S ####
 
 ## new infections
-n_S_progress[, ] <- rbinom(S[i, j], p_SE[i, j])
+
+## Compute the new infections with multiple strains using nested binomials
+n_S_progress[, 1, ] <- rbinom(S[i, k], p_SE[i, 1, k])
+n_S_progress[, 2:n_strains, ] <-
+ rbinom(S[i, k] - sum(n_S_progress[i, 1:(j - 1), k]), p_SE[i, j, k])
+
```
edknock, line 182: From what I understand, `p_SE[i, j, k]` represents the probability of being infected with strain `j`, but these are not mutually exclusive? And to me here it looks like we are giving automatic priority to strains with lower j.

Suppose we have three strains such that 
`p_SE[i,1,k] = 0.25`
`p_SE[i,2,k] = 0.5`
`p_SE[i,3,k] = 0.75`

As I understand the formulation we have, strain 1 would infect about 25%, then strain 2 would infect about 50% of 75% = 37.5%, and strain 3 would infect about 75% of 37.5% = 28.1%. This despite strain 3 having a greater force of infection. Do we want this or do we want something that would give more proportional prioritisation?

edknock: Would it be better to work out first the total number infected by all strains combined, and then determine how many were infected by each strain?

annecori: Fair point. I think with a small enough time step we should never get to those high probabilities but you are right that this is not ideal. 
@richfitz  this actually means the example I gave in the multinomial odin issue is not good. 
@edknock  what is your suggestion to do this properly? draw the total number of infections and then split them proportionally between strains?

richfitz: That sounds like drawing samples without replacement which is the multivariate hypergeometric - I'd rather not implement that here yet as it's really awkward

annecori: I think binomial followed by multinomial should work
I will try to implement tonight and get you to review

richfitz: OK great! I will see if I get time to make multinomial easier soon

annecori: I tried to fix this - @edknock if you could check you're happy with this that would be great

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# sircovid 0.9.0
+
+* Multistrain
```
richfitz, line 3: Can this please be expanded

R/carehomes_rt.R
```diff
@@ -47,7 +49,8 @@ carehomes_Rt <- function(step, S, p) {
       mean_duration_weighted <- drop(mean_duration[, , t])
     }
 
-    ngm <- outer(mean_duration_weighted, S_weighted) * m
+    ## In a multistrain model R0 is the max of R0 across strains
+    ngm <- outer(mean_duration_weighted, S_weighted) * m * max_strain_multiplier
```
richfitz, line 53: Can you open a [new issue](https://github.com/mrc-ide/sircovid/issues/new) to document what needs doing to make this robust

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: sircovid
 Title: SIR Model for COVID-19
-Version: 0.8.2
+Version: 0.9.0
```
richfitz, line 3: Can you increase the dust requirement to 0.5.7

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -0,0 +1,567 @@
+context("carehomes (multistrain)")
```
richfitz, line 1: I will do a quick style fix on this file

richfitz (changes requested):
Minor changes otherwise good to go I think.

It would be useful to know if a short run on the spim task works with this or if we need to update the indexes on anything exported

DESCRIPTION
richfitz: nm, I've done this

R/carehomes_rt.R
annecori: new issue here: https://github.com/mrc-ide/sircovid/issues/184

richfitz (approved):

PR #179, opened 2020-12-24
Explicit cast to avoid compiler warning

### LINKED ISSUE #170: Eliminate signedness warning
Harmless and can be taken care of by converting `step` into a signed integer before comparisons that use it

richfitz:
Finally closes our longstanding (but harmless) compiler warning. This has got more annoying as the number of time-varying things has increased. Warnings looked like:

```
basic.cpp: In member function ‘void basic::update(size_t, const real_t*, dust::rng_state_t<double>&, basic::real_t*)’:
basic.cpp:345:25: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare]
  345 |     real_t beta = (step >= internal.dim_beta_step ? internal.beta_step[internal.dim_beta_step - 1] : internal.beta_step[step + 1 - 1]);
      |                    ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
```

which we fix here by explicitly casting step to an integer. Generating the new sircovid code requires https://github.com/mrc-ide/odin.dust/pull/44 (and this PR is generated with code from that odin.dust PR).

Closes #170


inst/odin/basic.R
```diff
@@ -197,7 +197,7 @@ dim(beta_step) <- user()
 ## supported by odin (it could be made to support this). This code
 ## does currently create a compiler warning with -Wsign-compare on
 ## because we have an unsigned/signed integer comparison
```
johnlees, line 199: Can remove this, as warning now fixed!

johnlees (changes requested):
Looks sensible. Couldn't we do the cast just once though, rather than every time step is used?

richfitz:
> Looks sensible. Couldn't we do the cast just once though, rather than every time step is used?

Correct me if I am wrong but a static cast doesn't do any coersion it just tells the compiler that we're ok with the comparison here. With the current set-up in odin we can't do the assignment out of the loop though...

johnlees:
I wasn't sure what assembly actually results from these casts and looked it up:
- Pointer to pointer casts are no-op
- If you did `static_cast<int>` on a float type, this would require a call to `__ftoi`, plus a couple of moves
- unsigned to signed also causes a conversion, but this can be done with an addition and move to deal with the twos complement (so incredibly minor)

The CUDA guide advises us to use unsigned integers for loops (as apparently the overflow is better defined and easier/quicker to deal with). So maybe we should just use uints?

johnlees:
This is very minor of course, so I'll merge this an raise as in issue in dust

johnlees (approved):

PR #181, opened 2021-01-01
Add helper functions for future scenarios

### LINKED ISSUE #180: Add support for future Rt scenario setting

richfitz:
Adds helper functions for future scenarios:

* `add_future_betas` and `future_Rt` for describing future values of beta based on Rt
* `add_trajectory_incidence` and `drop_trajectory_incidence`: add and remove incidence trajectories
* `drop_trajectory_predicted` Drop predictions

Fixes #180


R/predict-support.R
```diff
@@ -0,0 +1,188 @@
...
+  assert_is(sample, "mcstate_pmcmc")
+  assert_is(rt, "Rt_trajectories")
+  sample <- drop_trajectory_predicted(sample)
+
+  ## We never want the future projections here, so exclude them
+  n_pars <- nrow(sample$pars)
+  n_time <- length(sample$trajectories$step)
+
+  ## Check that our trajectories are consistent with the sample:
+  i <- match(sample$trajectories$step, rt$step[, 1])
```
lwhittles, line 53: This check is causing problems when I integrate it into the task, specifically we have
>   range(sample$trajectories$date)
[1]   1 389
>   range(sample_Rt$date)
[1]  17 389 ` 

because `sample_Rt` starts from the date of importation, whereas `sample` always starts from 2020-01-01, regardless of the date of importation (with corresponding NA entries in state). 
There are a few ways we could fix this:
1. change the Rt calc to match the output of sample (a bit of a pain as we would then need full pmcmc reruns to do anything downstream) 
2. remove the pre-importation dates (and states) from sample$trajectories, which is probably worth doing in another helper function a là `drop_trajectory_predicted()`
3. something I haven't thought of

lwhittles (changes requested):

lwhittles (approved):
Tested thoroughly in the spi-m task and all seems to be working perfectly.

PR #185, opened 2021-01-06
Allow subsets of Rt to be computed, interpolating optionally

richfitz:
TODO:

- [ ] Fix the docs
- [x] Add more checks for sensible outputs
- [x] Use a reasonable default for min


R/carehomes_rt.R
```diff
@@ -115,7 +139,7 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL, type = NULL) {
         i_gen <- seq_len(nrow(ngm))[-c(i_CHW, i_CHR)]
         ngm <- ngm[i_gen, i_gen]
       }
-      ev <- eigen(ngm, FALSE, TRUE)$values
+      ev <- ev <- eigen(ngm, symmetric = FALSE, only.values = TRUE)$values
```
richfitz, line 142:
```suggestion
      ev <- eigen(ngm, symmetric = FALSE, only.values = TRUE)$values
```

richfitz (commented):
Can you update the version to `0.9.19` (about to merge .18 with unrelated changes from #212). We can sort out the interface later

annecori (approved):

PR #188, opened 2021-01-08
Expose dt/steps_per_day control

### LINKED ISSUE #187: Expose dt, and confirm working if this is changed

richfitz:
Just the core parameter handling part done here. Someone needs to check that the epi looks ok.

We should expand the testing to cover things like Rt but I wonder if this might actually not be that bad

Fixes #187

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: sircovid
 Title: SIR Model for COVID-19
-Version: 0.9.6
+Version: 0.9.7
 Authors@R:
     c(person(given = "Marc",
            family = "Baguelin",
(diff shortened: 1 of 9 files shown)


annecori:
I added some tests but the Rt test is failing: https://github.com/mrc-ide/sircovid/blob/50e118c15d08d890a1a7154706bf508f4fe26aee/tests/testthat/test-carehomes-rt.R#L147
Maybe @edknock you could look at this?

annecori:
Update on the above. I implemented a function to update the gamma parameters before running the model, so that the resulting mean duration of the compartments using the discretised gamma would match the mean of the continuous gamma distribution. This has fixed the Rt calculations - which are now similar across different dt values, but generates substantially different epidemic dynamics - e.g. attack rate is very different with different rt values. 
I think this means we need to rethink the calculation of Rt - I'm not sure it is correct to calculate the Rt using the discretised mean duration - maybe we should just use the mean of the continuous distribution? @edknock  would be great if you can have a think about this.

PR #191, opened 2021-01-13
Add multistrain to likelihood

edknock:
This adds the ability to fit to data for multistrains, using a binomial likelihood. We only use data for the over 25s.


richfitz (approved):
Happy with this from a code point of view, but this obviously should be checked by someone on the epi side

I am going to need to sort out why our computers reorder all the odin output though - I thought that I'd set that up to be locale independent but it's not working.

MJomaba (approved):
Looking good.

PR #195, opened 2021-01-16
Compute daily values from model, simplifying compare

richfitz:
This lays the groundwork for moving to a compiled comparison
function, which will not have access to the previous state, and
also allows some simplifications to mcstate.

Fixes #194

PR #199, opened 2021-01-20
Update compare function to work with mcstate 0.4.0

### LINKED ISSUE #198: Update to work with mcstate 0.4.0
Breaking changes as introduced in https://github.com/mrc-ide/mcstate/pull/96

richfitz:
This removes the "previous state" argument, unused since #195 and removed from mcstate in https://github.com/mrc-ide/mcstate/pull/96

Fixes #198


DESCRIPTION
```diff
@@ -48,7 +48,7 @@ Language: en-GB
 Depends: R (>= 3.1.0)
 Imports:
     dust (>= 0.6.2),
```
johnlees, line 50:
```suggestion
    dust (>= 0.6.5),
```

johnlees: May not be necessary?

johnlees (approved):

DESCRIPTION
richfitz: that's just changes to examples, so should be fine (though it will end up being installed about that version I think due to mcstate depending on it)

PR #201, opened 2021-01-20
Incorporate IFR calculation into the package

edknock:
Incorporates IFR/IHR calculation for the carehomes model. The weighting at time t is given by the expected infections at the next time step by group and vaccine class given pathway probabilities and the level of susceptibles and infectives at that time point.

Also calculates a "no vaccination" version, which ignores vaccine effects.

Currently the multistrain aspect of the model has zero direct impact on IFR/IHR


richfitz (changes requested):
This looks ok aside from the duplication with the Rt calculation which I can look at later. Obviously I can't check the science but if this is a port of what is in the tasks then I think we can go with that.

I would remove the persistant data, but if you're not immediately sure how, just add that to the issue #189 

Please create an issue and assign it to me about the duplication.

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# sircovid 0.9.15
```
richfitz, line 1: With apologies, we're going to be at .17 for this now!

R/carehomes_ifr_t.R
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+
+  t <- seq_along(step)
+  expected_infections_vacc <- vapply(t, calculate_expected_infections,
+                                     numeric(dim(S)[1]), no_vacc = FALSE)
+  expected_infections_no_vacc <- vapply(t, calculate_expected_infections,
+                                        numeric(dim(S)[1]), no_vacc = TRUE)
+
+  IFR_t_general <- vnapply(t, calculate_weighted_ratio,
```
richfitz, line 109: You might look here at the update that I added for this problem in the Rt calculation. Basically define a list of the options that you're changing (here drop_carehomes, no_vacc and IFR) then loop over the options. That will help with the duplication - see https://github.com/mrc-ide/sircovid/pull/208/files (you don't need to expose the "type" interface though you might choose to)

richfitz (changes requested):
I've got one substantive change on code structure, the rest I can see how it relates to the Rt stuff and I can try and harmonise that later

richfitz (approved):

PR #203, opened 2021-01-21
Fix multistrain Rt

edknock:
For the multistrain version of the model, `Rt` is calculated using `prob_strain` - which is essentially the proportion of force of infection that is the non-variant strain. However, if there is no force of infection then we yield `NA` values for `prob_strain`. This resulted in the eigenvalue calculation within the `Rt` calculation throwing an error.

This PR fixes this by setting `Rt` to `NA` in cases where `prob_strain` is `NA`.


R/carehomes_rt.R
```diff
@@ -80,25 +80,30 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL) {
...
+        i_CHR <- seq(p$n_groups, nrow(ngm), by = p$n_groups)
+        i_CHW <- i_CHR - 1
+        i_gen <- seq_len(nrow(ngm))[-c(i_CHW, i_CHR)]
+        ngm <- ngm[i_gen, i_gen]
+      }
+      ev <- eigen(ngm)$values
+      ev[Im(ev) != 0] <- NA
+      ev <- as.numeric(ev)
+      out <- max(ev, na.rm = TRUE)
     }
```
richfitz, line 106:
```suggestion
      out <- max(ev, na.rm = TRUE)
    }
    out
```

richfitz: this makes the return more explicit (currently the upper branch of the `if` is a long way from the bottom of the function)

richfitz (changes requested):
one tiny tweak here, but this seems reasonable

R/carehomes_rt.R
edknock: I meant to do that but clearly missed it. Thanks for catching that.

richfitz (approved):

PR #204, opened 2021-01-21
Setup for vaccine fitting

edknock:
This allows the vaccine daily doses to vary over time.

This uses a piecewise constant function I've introduced `sircovid_parameters_piecewise_constant`. Similar in construction to `sircovid_parameters_beta`. That function is a piecewise linear function, and could do with sensible renaming (issue #205).


R/parameters.R
```diff
@@ -97,6 +97,87 @@ sircovid_parameters_beta <- function(date, value, dt) {
...
+  }
+  assert_sircovid_date(date)
+  assert_increasing(date)
+  if (!is.null(date)) {
+    if (date[1L] != 0) {
+      stop("As 'date' is not NULL, first date should be 0")
+    }
+  }
+
+  f <- stats::stepfun(date, c(NA, value))
```
richfitz, line 178: It'll be equivalent (in fact it definitely is) but you can avoid creating the function here if you use

```
stats::approx(date, value, method = "constant", xout = seq(0, date[[length(date)]], by = dt)
```

(with probably some faff with the edge cases)

richfitz (changes requested):
Thanks for sorting this Ed. There will be changes made to projections tasks but no fitting tasks I believe. The projections tasks will need updating (and want updating) anyway to deal with the less lumpy vaccination.

I still think we need to deal with the issue with `model_pcr_and_serology_user`. For example in tests-carehomes-vaccination.R we have things like:

```
test_that("Vaccine progression through 3 classes works for susceptibles", {
  skip("TODO: can't vaccinate fast enough")
  ## Tests that:
  ## Every susceptible moves to waning immunity stage and stays there if
  ## everyone quickly gets vaccinated and loses immunity
  p <- carehomes_parameters(0, "england",
                            beta_value = 0,
                            rel_susceptibility = c(1, 0, 0),
                            rel_p_sympt = c(1, 1, 1),
                            rel_p_hosp_if_sympt = c(1, 1, 1),
                            vaccine_progression_rate = c(0, Inf, 0),
                            vaccine_daily_doses = Inf)
  ## TODO: Anne to look at tidying this parameter up:
  p$model_pcr_and_serology_user <- 0

  mod <- carehomes$new(p, 0, 1)
  info <- mod$info()
  mod$set_state(carehomes_initial(info, 1, p)$state)
  mod$set_index(integer(0))
  i <- 4:carehomes_n_groups()
  y <- mod$transform_variables(drop(
    dust::dust_iterate(mod, seq(0, 400, by = 4))))
  expect_equal(y$S[i, 1, 1], y$S[i, 3, 2])
  expect_equal(y$S[i, , 101], y$S[i, , 2])
})
```

This test is skipped at the moment because the model can't do infinite rates of vaccination. But there is some problem with who gets vaccinated when that is set to 1. @annecori  may remember the details.

This will have a big problem for us in fitting if we need serology turned on for the fits too, I think, as we will not end up seeing all the CHR/80+ groups be able to be vaccinated

R/parameters.R
edknock: Made this change

annecori (commented):
looks good to me with the following caveats:
- adding some examples to the carehomes_parameters would be good
- use explicit parameter names in examples for `sircovid_parameters_piecewise_constant` (and also for `sircovid_parameters_beta` actually) - in particular for `dt`
- `sircovid_parameters_beta` should probably be renamed (but could be done in a subsequent PR)
- what does this mean for ncov-outputs tasks? if any changes are needed there could we have a paired PR on ncov-outputs?

edknock:
> * `sircovid_parameters_beta` should probably be renamed (but could be done in a subsequent PR)

Already set an issue for this!

richfitz (approved):

PR #206, opened 2021-01-25
First working version of compiled compare

### LINKED ISSUE #200: Add a compiled comparison function

richfitz:
This requires very recent dust and odin.dust to compile but does the trick. It does not make things any faster at present but I think it's worth merging in in any case because (a) it was very fiddly to write and (b) it does open up some additional optimisations later (and I'd like to try this out on a windows PC at some point).

Very small effect on downstream tasks, entire interface is exactly the same and the R compare functions still work. Tests provided for validation with the R ones so when one compare function is expanded we should reflect that in the other.

This is failing on windows with different answers being produced somewhat unexpectedly. I will investigate before this is merged, but I think that this is otherwise OK for a look now.

* one of: @edknock, @MJomaba, @annecori, @lwhittles - is the .cpp compare function clear enough? Do you think you could hack something in if I was on leave? What more docs would you like?
* @johnlees: can you look over the .cpp in particular and see what you think?

**BREAKING CHANGES**

* ~The data element formerly known as `new` must now be called `new_conf`. This is because `new` is a reserved word in C++~ (Done more thoroughly in #212)
* Parameters have been reordered, so anything looking in `$observation$x` should look for `$x`

Note that there is a test failure which looks like an unexplained platform dependent issue. I will investigate that once I know how this will fly with the science side (update: I cannot replicate this on any windows machine - I get the same numbers as linux so must be some library issue?)

Fixes #200


inst/odin/compare_basic.cpp
```diff
@@ -0,0 +1,30 @@
+#include <dust/densities.hpp>
```
johnlees, line 1: Have we been using headers in included snippets? i.e. does the vaccine code use dust header files too?
Might be difficult to remember/know which one to pick for the non-dust developer, so we could either just make sure this is documented, or we could presumably add this header in to dust.hpp like we do with the distributions.

inst/odin/compare_carehomes.cpp
```diff
@@ -0,0 +1,215 @@
+#include <dust/densities.hpp>
+
+template <typename real_t>
+real_t ll_nbinom(real_t data, real_t model, real_t kappa, real_t exp_noise,
+                 dust::rng_state_t<real_t>& rng_state) {
```
johnlees, line 5: Nice function overloading here 👍

inst/odin/compare_carehomes.cpp
```diff
@@ -0,0 +1,215 @@
...
+  return (sensitivity * pos + (1 - specificity) * neg) / (pos + neg);
+}
+
+// Lots of data items here. Even though some of these are counts (in
+// fact, most of them are counts) we'll pull them in as doubles
+// because that will help with the NA handling. Once this is working
+// reliably we could cast these explicitly as ints, but there is not
+// likely to be any major gain here.
+//
+// [[odin.dust::compare_data(icu = double)]]
```
johnlees, line 51: Is there any way to have these as real_t, or turned into the right type by odin.dust?

inst/odin/compare_carehomes.cpp
```diff
@@ -0,0 +1,215 @@
...
+// [[odin.dust::compare_function]]
+template <typename T>
+typename T::real_t compare(const typename T::real_t * state,
+                           const typename T::data_t& data,
+                           const typename T::internal_t internal,
+                           std::shared_ptr<const typename T::shared_t> shared,
+                           dust::rng_state_t<typename T::real_t>& rng_state) {
+  typedef typename T::real_t real_t;
+
+  // State variables; these largely correspond to the quantities in data
```
johnlees, line 81: Why are some unpacked here, and some unpacked where they are used?

inst/odin/compare_carehomes.cpp
```diff
@@ -0,0 +1,215 @@
...
+  const real_t model_strain_over25_prob_pos =
+    test_prob_pos<real_t>(model_sympt_cases_non_variant_over25,
+                          model_sympt_cases_over25 -
+                          model_sympt_cases_non_variant_over25,
+                          1, 1, exp_noise,
+                          rng_state);
+
+  // Note that in ll_nbinom, the purpose of exp_noise is to allow a
+  // non-zero probability when the model value is 0 and the observed
+  // value is non-zero (i.e. there is overreporting)
```
johnlees, line 150: Is that necessarily over-reporting? Couldn't the model be wrong?

johnlees (changes requested):
I've just looked at the code part, not whether the compare function is correct

inst/odin/compare_carehomes.cpp
richfitz: That just mimics how the original compare function was written to make the connection with the R easier https://github.com/mrc-ide/sircovid/blob/i200-compiled-compare/R/carehomes.R#L514-L527

inst/odin/compare_carehomes.cpp
richfitz: No idea! This is ported over from the R :) https://github.com/mrc-ide/sircovid/blob/i200-compiled-compare/R/carehomes.R#L569-L571

inst/odin/compare_carehomes.cpp
richfitz: Good idea, though it will require odin.dust support: added as https://github.com/mrc-ide/odin.dust/issues/57

inst/odin/compare_basic.cpp
richfitz: We've managed to avoid the snippets approach to this point, but yeah, this is something that we should be able to handle ourselves. The tidiest way would be for odin.dust or dust to include this for us I think https://github.com/mrc-ide/odin.dust/issues/58

richfitz: These headers now come directly from dust so are removed here; I think this is a good call

inst/odin/compare_carehomes.cpp
```diff
@@ -0,0 +1,213 @@
...
+    test_prob_pos(model_sero_pos,
+                  odin(N_tot_15_64) - model_sero_pos,
+                  odin(sero_sensitivity),
+                  odin(sero_specificity),
+                  exp_noise,
+                  rng_state);
+
+  // Strain
+  const real_t model_strain_over25_prob_pos =
+    test_prob_pos<real_t>(model_sympt_cases_non_variant_over25,
```
edknock, line 140: Why do we have `test_prob_pos<real_t>` here but not for the other uses of `test_prob_pos`?

richfitz: That was needed to disambiguate the template to use as the compiler found it ambiguous due to sensitivity and specificity being untyped. I've updated that to be more obvious now

docker/Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN install2.r --error \
 
 RUN Rscript -e 'remotes::install_github(c("mrc-ide/dust", "mrc-ide/mcstate"))'
 
-RUN Rscript -e 'remotes::install_github("mrc-ide/dust")'
+RUN Rscript -e 'remotes::install_github("mrc-ide/dust@v0.7.5")'
```
johnlees, line 24: Should this version be pinned?

johnlees (approved):

docker/Dockerfile
richfitz: not any more!

richfitz: oh, hang on that's there to invalidate the docker build cache. I'll do something else that will be less annoying in future

PR #213, opened 2021-02-11
Restructure symptomatic pathway to incorporate presymptomatic compartment and two symptomatic compartments

edknock:
- Restructures the symptomatic flow in the carehomes model so that `I_C` is replaced by `I_P -> I_C_1 -> I_C_2`

- `sympt_cases` for use with pillar 2 data now comes from individuals moving from `I_C_1` to `I_C_2`

- individuals in `I_P` can be vaccinated and change vaccination stage 

- Added transmission parameters for `I_A`, `I_P`, `I_C_1` and `I_C_2` so levels of relative transmissibility (including potentially zero transmissibility) in these compartments can be easily altered


inst/odin/carehomes.R
```diff
@@ -1303,19 +1410,18 @@ initial(sero_pos) <- 0
 update(sero_pos) <- sum(new_T_sero_pos[4:13, , , ])
 
 initial(cum_sympt_cases) <- 0
-new_sympt_cases <- sum(n_EI_C) + sum(n_EI_C_next_vacc_class)
+new_sympt_cases <- sum(n_I_C_1_progress[, , k_C_1, ])
```
annecori, line 1413: should this use n_I_P_progress instead?

inst/odin/carehomes.R
```diff
@@ -1303,19 +1410,18 @@ initial(sero_pos) <- 0
...
-  sum(n_EI_C_next_vacc_class[6:n_groups, , ])
+new_sympt_cases_over25 <- sum(n_I_C_1_progress[6:n_groups, , k_C_1, ])
 update(cum_sympt_cases_over25) <- cum_sympt_cases_over25 +
   new_sympt_cases_over25
 
 initial(cum_sympt_cases_non_variant_over25) <- 0
-new_sympt_cases_non_variant_over25 <- sum(n_EI_C[6:n_groups, 1, ]) +
-  sum(n_EI_C_next_vacc_class[6:n_groups, 1, ])
+new_sympt_cases_non_variant_over25 <-
+  sum(n_I_C_1_progress[6:n_groups, 1, k_C_1, ])
```
annecori, line 1424: should this use n_I_P_progress instead?

annecori (commented):
This looks ok to me minus the following:
- I think we should define the transmissibility of I_A, I_P, I_C_1 and I_C_2 relative to one of those, as I think you would not want to fix all of these to sth != 1 (minor)
- you could also merge IC_1 and IC_2 in to a single array since you have k_I_C sequential compartments. Then you'd need to define the transmissibility of I_C as an array as well rather than a scalar but may simplify the code? (minor)
- I think the calculations of new symptomatic is wrong as it uses I_C_1_progress but I believe it should use I_P_progress? I've tagged the specific places

inst/odin/compare_carehomes.cpp
```diff
@@ -158,18 +161,30 @@ typename T::real_t compare(const typename T::real_t * state,
     ll_nbinom(data.hosp, odin(phi_hosp) * model_hosp,
               odin(kappa_hosp), exp_noise, rng_state);
 
-  // We will either compute ll_deaths_hosp and ll_deaths_comm *or* we
-  // will compute the combined version.
+  // We will compute one of the following:
```
richfitz, line 164: nice to see how you can port this over here; hope it wasn't too faffy

tests/testthat/test-basic.R
```diff
@@ -71,7 +71,7 @@ test_that("compiled compare function is correct", {
                      start_date, pars$dt)
 
   np <- 10
-  mod <- basic$new(pars, 0, np, seed = 1L)
+  mod <- basic$new(pars, 0, np, seed = 3L)
```
richfitz, line 74: why was this change needed?

tests/testthat/test-carehomes-vaccination.R
```diff
@@ -274,6 +276,56 @@ test_that("Vaccination of asymptomatic infectious individuals works", {
 })
 
 
+test_that("Vaccination of presymptomatic infectious individuals works", {
+  skip("TODO: can't vaccinate fast enough")
```
richfitz, line 280: was this moved from somewhere? We shouldn't add new tests that don't run (we should find time to fix the existing skipped ones at some point)

richfitz (changes requested):
I can't interpret codecov's report- it might fix up once all tests pass again.

Please:
* merge origin/master into this PR as Anne has done some updates with the most recent version of packages
* You should then upgrade both dust and odin dust to the versions on github (not drat) as there are quite a few changes there
* You need to tone town the exact equality check that is failing on windows to something like abs(diff) < eps for eps ~ 1e-6

tests/testthat/test-carehomes-vaccination.R
edknock: I copied and adapted the other vaccination tests. I think there's 12 total now that are being skipped. Yes, we should fix them.

tests/testthat/test-basic.R
edknock: I should've highlighted that this was temporary. With `seed = 1L` we were seeing a corner case appear that had differences between `dnbinom` and the dust `dnbinom`, so the two compares weren't matching. You fixed this though, so can revert this change now.

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -580,7 +580,7 @@ test_that("Can calculate Rt with a second less infectious variant", {
 test_that("Can calculate Rt with a second more infectious variant", {
   ## Seed with 10 cases on same day as other variant
   p <- carehomes_parameters(sircovid_date("2020-02-07"), "england",
-                            strain_transmission = c(1, 10),
+                            strain_transmission = c(1, 5),
```
edknock, line 583: With 10, this now causes the epidemic to die out within this time frame. And once we have 0 infectives, the multistrain version yields NA for R, so the test would fail. Hence I have reduced it to 5, so that the epidemic doesn't die out within the time frame.

annecori:
> This looks ok to me minus the following:
> 
> * I think we should define the transmissibility of I_A, I_P, I_C_1 and I_C_2 relative to one of those, as I think you would not want to fix all of these to sth != 1 (minor)
> * you could also merge IC_1 and IC_2 in to a single array since you have k_I_C sequential compartments. Then you'd need to define the transmissibility of I_C as an array as well rather than a scalar but may simplify the code? (minor)
> * I think the calculations of new symptomatic is wrong as it uses I_C_1_progress but I believe it should use I_P_progress? I've tagged the specific places

The last bullet point has been addressed so happy with that. 
I still think the first bullet point should be addressed otherwise we're overparameterising the model. 
Second bullet point is only a suggestion as we've tried to avoid having hardcoded indexes. Not essential but would be good to hear your thoughts @edknock

edknock:
> > This looks ok to me minus the following:
> > 
> > * I think we should define the transmissibility of I_A, I_P, I_C_1 and I_C_2 relative to one of those, as I think you would not want to fix all of these to sth != 1 (minor)
> > * you could also merge IC_1 and IC_2 in to a single array since you have k_I_C sequential compartments. Then you'd need to define the transmissibility of I_C as an array as well rather than a scalar but may simplify the code? (minor)
> > * I think the calculations of new symptomatic is wrong as it uses I_C_1_progress but I believe it should use I_P_progress? I've tagged the specific places
> 
> The last bullet point has been addressed so happy with that.
> I still think the first bullet point should be addressed otherwise we're overparameterising the model.
> Second bullet point is only a suggestion as we've tried to avoid having hardcoded indexes. Not essential but would be good to hear your thoughts @edknock

- re merging I_C_1 and I_C_2: I would prefer to think of them as separate compartments, as they have different numbers of stages, rates, levels of transmissibility etc. If anything, I'd rather rename them if we potentially get better names.

- re the transmission parameters, I can set them up as relative to I_C_1 if you think that's the best idea, though then that hardcodes transmissibility being relative to I_C_1, so we can't define transmissibility as being relative to a different compartment then. We also would not be able to have zero transmissibility in I_C_1 if we wanted that. So we would lose flexibility - but perhaps this is flexibility we would never actually use anyway?

annecori:
ok let's leave as is for now then @edknock. Perhaps in carehomes_parameters we could add a check that one of the transmissibility parameters is 1? just a suggestion.

edknock:
> ok let's leave as is for now then @edknock. Perhaps in carehomes_parameters we could add a check that one of the transmissibility parameters is 1? just a suggestion.

These values are hardcoded into the function, so not sure the check is necessary. Worth noting that you could get around such a check by changing the values after you've run carehomes_parameters

R/carehomes.R
```diff
@@ -404,6 +404,12 @@ carehomes_parameters <- function(start_date, region,
   ## Proportion of population with covid-like symptoms without covid
   ret$p_NC <- p_NC
 
+  ## relative transmissibility of various I compartments
+  ret$I_A_transmission <- 0.363
```
richfitz, line 408: where has this number come from? can we add a reference as we will forget?

inst/extdata/population.csv
```diff
@@ -1,18 +1,18 @@
...
-"40",4019539,3414297,169556,319432,116254,399928,475129,603274,402791,646347,551341,311774
-"45",4402122,3715812,199072,361458,125780,473672,572987,713583,448631,596141,620052,374176
-"50",4661015,3907461,220480,401090,131984,494502,605129,739225,464715,564205,644651,403024
-"55",4405908,3670651,217480,393123,124654,460456,573996,675572,421265,479470,582461,380109
-"60",3755185,3111835,192853,344693,105804,391498,501802,579975,362595,378068,491505,338931
-"65",3368199,2796740,182142,299444,89873,363640,464318,551557,341872,309586,458710,332910
-"70",3318867,2779326,179286,278856,81399,347142,440635,533581,341711,265847,462533,333351
-"75",2325296,1940686,126785,195951,61874,239787,298198,368463,227741,189753,313538,225646
-"80",3362599,2836964,174163,269894,81578,337891,433619,526725,349658,294027,487502,339312
+"age","uk","england","wales","scotland","northern_ireland","east_of_england","london","midlands","north_east_and_yorkshire","north_west","south_east","south_west"
```
richfitz, line 1: Can you update the README in the inst/extdir directory indicating where these numbers have come from, probably leaving a reference to the old version too

richfitz (changes requested):
Two small data provenance requests

R/carehomes.R
edknock: This number has been calculated by us - which will be described in the revised paper.

richfitz (approved):

PR #215, opened 2021-02-12
Faster Rt calculation

### LINKED ISSUE #209: Faster leading eigenvalue calculation, again

richfitz:
It depends a lot on tuning and the situation, but this is a bunch faster than the previous version (at least 5x and possibly as much as 50x)

The bookkeeping within the Rt calculation has changed considerably; we now treat the entire time series at the same time and then compute the leading eigenvalue across that - the eigen1 package helps here.

I've kept the interpolation bit in but be careful as it's not been thoroughly tested. It's a different approach to before, but it should work.

Note that the eigenvalue calculations are not deterministic (they use a stochastic starting vector for the first entry and then will be influenced by the previous time point).

~Before merging https://github.com/mrc-ide/eigen1/pull/1 needs merging and the branch pin here updating~

Fixes #209


tests/testthat/test-carehomes-multistrain.R
```diff
@@ -685,3 +686,41 @@ test_that("If prob_strain is NA then Rt is NA ", {
...
+  reduced_susceptibility <- 0.2 # can put anything <1 here
+  p <- carehomes_parameters(sircovid_date("2020-02-07"), "england",
+                            strain_transmission = c(1, 0.1),
+                            strain_seed_date =
+                              rep(sircovid_date("2020-02-07"), 2),
+                            strain_seed_value = 10,
+                            rel_susceptibility = c(1, reduced_susceptibility),
+                            rel_p_sympt = c(1, 1),
+                            rel_p_hosp_if_sympt = c(1, 1),
+                            waning_rate = 1 / 20)
```
edknock, line 707: Possibly want doses specified here? Otherwise I think there's no actual vaccination going on in this simulation

edknock (approved):
Tested and it works well

PR #220, opened 2021-02-18
removing calculation of probabilities from rate from probabilities

annecori:
This avoids useless calculation. 
If p is the proportion to vaccinate in a given time step and r is the rate of vaccination, we were doing:
- in the cpp function vaccination_schedule: calculate p, then calculate r = -log(1-p) / dt, return r
- in the odin code: use r from function vaccination_schedule code, then compute p = 1 - exp(-r*dt)

This has been changed so that: 
- in the cpp function vaccination_schedule: calculate p, return p
- in the odin code: use p from cpp function vaccination_schedule


richfitz (approved):

richfitz (changes requested):
ugh, sorry, I thought I'd seen the tests passing - will investigate later

richfitz (approved):

PR #238, opened 2021-03-11
EpiEstim

annecori:
new functions to compute EpiEstim Rt - tried to align as much as possible with the equivalent functions computing Rt from NGM in terms of input and output format


R/carehomes_rt.R
```diff
@@ -224,6 +224,199 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
...
+    ## but we don't care
+    R_i <- suppressWarnings(EpiEstim::estimate_R(incid = incidence[i, ],
+                                                 config = config)$R)
+
+    R_i_shape_scale <-
+      epitrix::gamma_mucv2shapescale(mu = R_i$`Mean(R)`,
+                                     cv = R_i$`Std(R)` / R_i$`Mean(R)`)
+
+    R_sample[, n_R * (i - 1) + seq_len(n_R)] <-
+      t(sapply(seq_len(length(t_start)), function(e) {
```
richfitz, line 388: Remove this sapply please. But also I would consider refactoring so that this is readable. Generally avoid inline definition of anonymous functions that require braces

R/carehomes_rt.R
```diff
@@ -224,6 +224,199 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
...
+  R_sample <- matrix(NA, length(t_start), n_R * np)
+
+  for (i in seq_len(np)) {
+    ## this function gives warning when precision in estimates is not good
+    ## but we don't care
+    R_i <- suppressWarnings(EpiEstim::estimate_R(incid = incidence[i, ],
+                                                 config = config)$R)
+
+    R_i_shape_scale <-
+      epitrix::gamma_mucv2shapescale(mu = R_i$`Mean(R)`,
```
richfitz, line 384: Do not use backquotes, use `R_i[["Mean(R)"]]`

NAMESPACE
```diff
@@ -44,9 +45,18 @@ export(vaccine_schedule_data_future)
 export(vaccine_schedule_from_data)
 export(vaccine_schedule_future)
 export(vaccine_schedule_scenario)
+importFrom(EpiEstim,estimate_R)
+importFrom(EpiEstim,make_config)
+importFrom(distcrete,distcrete)
+importFrom(epitrix,gamma_mucv2shapescale)
```
richfitz, line 51: Please implement this function directly in the package rather than adding as a dependency, as the definition looks trivial

R/carehomes_rt.R
```diff
@@ -224,6 +224,199 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
 }
 
 
+##' @title Performs draws from the generation time distribution
```
richfitz, line 227: Can you move all this code into a file called epiestim or similar

R/carehomes_rt.R
```diff
@@ -224,6 +224,199 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
...
+##'
+##' @param p A single [carehomes_parameters()] object.
+##'
+##' @param n The number of draws to perform
+##'
+##' @return a vector of `n` values of the generation time
+##'
+##' @importFrom distcrete distcrete
+##' @importFrom stats runif rpois
+draw_one_GT_sample <- function(p, n = 1000) {
```
richfitz, line 237: all function names must be lower case unless completely unavoidable

R/carehomes_rt.R
```diff
@@ -224,6 +224,199 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
 }
 
 
+##' @title Performs draws from the generation time distribution
+##'
+##' @param p A single [carehomes_parameters()] object.
+##'
+##' @param n The number of draws to perform
```
richfitz, line 231: we have not typically documented internal functions like this. If you do do this, please add `@nord` to prevent the creation of files in `man`. However, you do this inconsistently - did you mean to export this?

R/carehomes_rt.R
```diff
@@ -224,6 +224,199 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
...
+    ret <- list(t_start = time_start,
+                t_end = time_end,
+                Rt = R_sample,
+                Rt_summary = summary_R)
+  } else {
+    ret <- list(t_start = time_start,
+                t_end = time_end,
+                Rt_summary = summary_R)
+  }
+
```
richfitz, line 416: you forgot the `ret` here at the end. However, the formulation above is error prone. I would suggest:

```r
  ret <- list(t_start = time_start,
              t_end = time_end,
              Rt = R_sample,
              Rt_summary = summary_R)
  if (save_all) {
    ret$Rt <- R_sample
  }
  ret
```

I would also suggest renaming `save_all` with `all_times` (or some indication of what `all` is). you can drop the `save` part to keep the name under control

R/carehomes_rt.R
```diff
@@ -224,6 +224,199 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
...
+  # generate the config based on incidence for first parameter set
+  config <- EpiEstim::make_config(incid = incidence[1, ],
+                                  t_start = t_start,
+                                  t_end = t_end,
+                                  method = "non_parametric_SI",
+                                  si_distr = gt_distr,
+                                  mean_prior = mean_prior,
+                                  std_prior = sd_prior)
+
+  R_sample <- matrix(NA, length(t_start), n_R * np)
```
richfitz, line 375: NA_real_

R/carehomes_rt.R
```diff
@@ -224,6 +224,199 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
...
+carehomes_EpiEstim_Rt_trajectories <- function(step, incidence, p,
+                                  sliding_window_ndays = 7,
+                                  mean_prior = 1,
+                                  sd_prior = 1,
+                                  n_GT = 10000,
+                                  n_R = 1000,
+                                  save_all = TRUE) {
+  gt_distr <- draw_one_GT_distr(p = p, n = n_GT)
+
+  T <- ncol(incidence)
```
richfitz, line 361: I would not use `T` as a variable name

DESCRIPTION
```diff
@@ -50,7 +50,10 @@ Imports:
     dust (>= 0.8.1),
     eigen1,
     mcstate (>= 0.4.0),
-    socialmixr
+    socialmixr,
```
richfitz, line 53: dependencies should be sorted alphabetically (with upper case letters sorted before lower case). As discussed below you can and should remove epitrix.

I think that the remaining 2 belong in Suggests as they are not needed for core running of the package. You would remove the importFrom statements, which were not needed anyway as you were using the fully qualified name

R/carehomes_rt.R
```diff
@@ -224,6 +224,199 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
...
+##' of Rt for that sliding window (columns of the matrix)
+##' `Rt_summary` a matrix containing for each sliding window (each row in the
+##' matrix) the 2.5%, 50%, 97.5% quantiles and the mean of Rt for that sliding
+##' window (columns of the matrix)
+##'
+##' @export
+##' @importFrom EpiEstim make_config estimate_R
+##' @importFrom epitrix gamma_mucv2shapescale
+##' @importFrom stats quantile rgamma
+carehomes_EpiEstim_Rt_trajectories <- function(step, incidence, p,
```
richfitz, line 352: I think `carehomes_Rt_trajectories_EpiEstim` would be marginally better. However, there's no great name here, and the casing here irritates me a lot. I would personally prefer `carehomes_rt_trajectories_epiestim` with the package renamed

R/carehomes_rt.R
```diff
@@ -224,6 +224,199 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
...
+          ret <- rgamma(n_R, shape = R_i_shape_scale$shape[e],
+                        scale = R_i_shape_scale$scale[e])
+        } else {
+          ret <- rep(NA, n_R)
+        }
+        ret
+      }))
+  }
+
+  summary_R <- apply(R_sample, 1, quantile, c(0.025, 0.5, 0.975), na.rm = TRUE)
```
richfitz, line 399: these quantiles should be configurable

R/carehomes_rt.R
```diff
@@ -224,6 +224,199 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
...
+  mean_R <- apply(R_sample, 1, mean, na.rm = TRUE)
+  summary_R <- t(rbind(summary_R, mean_R))
```
richfitz, line 401: I am currently in the process of refactoring the Rt methods to return time in the last dimension to match the rest of the functions in the package. Could you make sure yours do that too from the start

R/support.R
```diff
@@ -362,17 +362,34 @@ combine_rt <- function(rt, samples) {
 }
 
 
+combine_rt_EpiEstim <- function(rt, samples) {
+  ## Ensure all trajectories are the same length
+  ret <- rt[[1L]]
+  what <- "Rt"
```
richfitz, line 368: what is going on here with `what`? you define it, then use the string that you would have defined? Seems like you might have been part way through tidying this up

R/support.R
```diff
@@ -362,17 +362,34 @@ combine_rt <- function(rt, samples) {
...
+  if (!("Rt" %in% names(ret))) {
+    stop("rt$Rt missing. Did you forget 'save_all = TRUE'
```
richfitz, line 370: Do not write multiline strings. Use paste here, as elsewhere in the package

```
    stop(paste("rt$Rt missing. Did you forget 'save_all = TRUE'",
               in 'carehomes_EpiEstim_Rt_trajectories'?"))
```

R/support.R
```diff
@@ -393,3 +410,43 @@ combine_rt1 <- function(what, rt, samples) {
 
   Reduce(`+`, ret)
 }
+
+
+combine_rt1_EpiEstim <- function(what, rt, samples) {
```
richfitz, line 415: See https://github.com/mrc-ide/sircovid/pull/227 for the changes I was making to these functions before everything got backed up on the vaccination work

tests/testthat/test-carehomes-rt.R
```diff
@@ -343,3 +344,213 @@ test_that("Parameters affect Rt as expected", {
   helper("k_G_D", 1, 2)
 
 })
+
+
+test_that("Can calculate EpiEstim Rt", {
+  p <- carehomes_parameters(sircovid_date("2020-02-07"), "england",
```
richfitz, line 350: Once epiestim etc has been moved to Suggests, this becomes

```
skip_if_not_installed("EpiEstim")
```

R/carehomes_rt.R
```diff
@@ -224,6 +224,199 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
...
+                                  method = "non_parametric_SI",
+                                  si_distr = gt_distr,
+                                  mean_prior = mean_prior,
+                                  std_prior = sd_prior)
+
+  R_sample <- matrix(NA, length(t_start), n_R * np)
+
+  for (i in seq_len(np)) {
+    ## this function gives warning when precision in estimates is not good
+    ## but we don't care
```
richfitz, line 379: I would not leave a comment here saying "we don't care" without explaining why we don't care.

richfitz (changes requested):
I can fix the buildkite in a bit

R/epiestim_rt.R
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  sampled_gt <- numeric(0)
+  i <- 0
+
+  while (i < n) {
+    ## Draw exposed period
+    sample_E <- draw_from(1, p$k_E, p$gamma_E)
+    ## Draw symptomatic vs asymptomatic path
+    path_C <- runif(1) <= p$p_C[[1]]
```
lwhittles:
```suggestion
    path_C <- runif(1) <= p$p_C
```
we already checked p$p_C is a scalar

R/epiestim_rt.R
```diff
@@ -0,0 +1,191 @@
...
+      sample_I <- draw_from(1, p$k_A, p$gamma_A)
+      infectivity <- p$I_A_transmission
+    } else {
+      sample_I <- draw_from(1, p$k_P, p$gamma_P) +
+        draw_from(1, p$k_C_1, p$gamma_C_1)
+      infectivity <- 1
+    }
+    n_secondary_cases <- rpois(1, lambda = infectivity * sample_I)
+    tmp_gt <- seq(sample_E + p$dt, sample_E + sample_I, p$dt)
+    sampled_gt <- c(sampled_gt,
```
lwhittles, line 45: Maybe worth putting in a check here that 
`length(tmp_gt) > 1` 
because if `sample_I == p$dt`, which I think is technically possible, then `sample(tmp_gt, ...)` will behave like `sample.int(sample_E, ...)` and all hell will break loose :)

lwhittles (changes requested):
This looks great to me, agree with the simulation algorithm, and while I'm not familiar with the EpiEstim interface this seems to be doing what it's supposed to. I've falgged a potential bug to splat

R/epiestim_rt.R
annecori: I think the check is that all the values in `p$p_C` are the same (`if (length(unique(p$p_C)) > 1`) but `p$p_C` is always a vector with 19 elements - so I think I should still use [[1]]

R/support.R
```diff
@@ -362,17 +362,34 @@ combine_rt <- function(rt, samples) {
 }
 
 
+combine_rt_EpiEstim <- function(rt, samples) {
```
richfitz, line 365: This function is never used so can be removed

R/support.R
richfitz, line 415: This function is never used so can be removed

R/utils.R
```diff
@@ -186,3 +186,10 @@ block_expand <- function(m, n) {
   len <- nrow(m) * n
   matrix(t(matrix(m, nrow(m), len)), len, len, byrow = TRUE)
 }
+
+
+gamma_mucv2shapescale <- function(mu, cv) {
+  shape <- 1 / (cv^2)
+  scale <- mu * cv^2
+  return(list(shape = shape, scale = scale))
```
richfitz, line 194: please drop the explicit `return`

richfitz (changes requested):

R/epiestim_rt.R
annecori: @lwhittles  can you please check if you're happy with my response to this and the fix (from @richfitz) to your other comment about the sample issue? Once you're happy can you pls approve changes? thx!

R/support.R
annecori: have added a test for this

R/support.R
annecori: have added a test for a function which calls this one

R/support.R
```diff
@@ -361,18 +361,53 @@ combine_rt <- function(rt, samples) {
...
+##'   regions (or whatever else you are aggregating on) based on the
+##'   parameters in `samples`.
+##'
+##' @export
+combine_rt_epiestim <- function(rt, samples, q = NULL) {
+  q <- q %||% c(0.025, 0.5, 0.975)
+  ## Ensure all trajectories are the same length
+  ret <- rt[[1L]]
+  if (!("Rt" %in% names(ret))) {
+    stop(paste("rt$Rt missing. Did you forget 'save_all_Rt_sample = TRUE'",
```
richfitz, line 387: last issue here - this needs a test that triggers the error

richfitz (changes requested):

R/support.R
annecori: done here https://github.com/mrc-ide/sircovid/pull/238/commits/1d29f403c4a1d04058b7eb4c55f88142921642ca

R/epiestim_rt.R
lwhittles: You're quite right!

lwhittles (approved):
Looks good!

richfitz (approved):

PR #240, opened 2021-03-12
support multiregion carehomes_particle_filter_data

RaphaelS1:
Allows for nested models in carehomes_particle_filter_data by checking the pillar2 stream requirement (one stream per region) by looping and checking regions individually.


MJomaba (approved):
Looking good

PR #243, opened 2021-03-13
Time varying gammas all

pabloperguz:
Support for time-varying parameters for all gammas for hospital progressions has been added. 
All but tests that use data/pmcmc.rds pass; not sure how to declare a zero value for gamma_H_D in that RDS file, which is why I believe it's failing


R/carehomes.R
```diff
@@ -340,6 +340,16 @@ carehomes_parameters <- function(start_date, region,
...
+  ## implementation of time-varying progression gammas
+  progression$gamma_H_R_step <- max(progression$gamma_H_R)
+  progression$gamma_W_R_step <- max(progression$gamma_W_R)
+  progression$gamma_ICU_W_R_step <- max(progression$gamma_ICU_W_R)
+  progression$gamma_H_D_step <- max(progression$gamma_H_D)
+  progression$gamma_W_D_step <- max(progression$gamma_W_D)
+  progression$gamma_ICU_W_D_step <- max(progression$gamma_ICU_W_D)
+  progression$gamma_ICU_D_step <- max(progression$gamma_ICU_D)
+  progression$gamma_ICU_pre_step <- max(progression$gamma_ICU_pre)
+
```
edknock, line 352: The above gamma_X are all scalar currently, so I'd remove the max()

DESCRIPTION
```diff
@@ -34,6 +34,10 @@ Authors@R:
            family = "Cori",
            role = c("aut"),
            email = "a.cori@imperial.ac.uk"))
```
edknock: change the second `)` to `,`

edknock (approved):

PR #246, opened 2021-03-15
add a rank argument to combine functions

annecori:


R/support.R
```diff
@@ -447,6 +453,7 @@ combine_rt1 <- function(what, rt, samples) {
     t(s$trajectories$state["infections_inc", , i]), samples, idx)
   rt_what <- Map(function(r, i) r[[what]][i, ], rt, idx)
 
+  if (rank) {
```
richfitz, line 456: please reindent

R/support.R
```diff
@@ -482,12 +493,13 @@ combine_rt1_epiestim <- function(what, rt, samples) {
...
   rt_what <- Map(function(r, i) r[[what]][, i], rt, idx)
 
-  ## Calculate rank of particles by area under Rt curve
-  ## average across sets of n_R_per_traj Rs
   n_R_per_traj <- round(nrow(rt_what[[1]]) / nrow(incidence[[1]]))
   incidence_rep <- lapply(incidence, function(m)
     m[rep(seq_len(nrow(m)), each = n_R_per_traj), ])
-  idx_incidence <- rep(seq_len(nrow(incidence[[1]])), each = n_R_per_traj)
+
+  if (rank) {
```
richfitz, line 500: reindent

tests/testthat/test-carehomes-pmcmc.R
```diff
@@ -157,6 +157,32 @@ test_that("can combine rt calculations over trajectories", {
   expect_equal(res$Rt_all, cmp$Rt_all)
 })
 
+
+test_that("can combine rt calculations over trajectories without reordering", {
```
richfitz, line 161: I think that the test missing here is that you can rank the trajectories and have the same trajectories out the end. So say you have trajectories {x1, x2, x3, x4} and you reorder them {x1, x3, x4, x2} then you'd check that `expect_equal(ordered[c(1, 4, 3, 2)], unordered)` say

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: sircovid
 Title: SIR Model for COVID-19
-Version: 0.10.9
+Version: 0.10.10
```
richfitz, line 3: .11 I think

tests/testthat/test-carehomes-pmcmc.R
richfitz: though given you're basically doing the opposite the test is easier here , but I think a more explicit test is worthwhile

richfitz (changes requested):

tests/testthat/test-carehomes-pmcmc.R
annecori: hopefully the new test "when combining rt calculations the output has the expected order" addresses that

richfitz (approved):

PR #255, opened 2021-03-19
Add GPU support

### LINKED ISSUE #237: Include gpu code in the package

richfitz:
This adds a helper function from which we can compile out a version of the model to run on the GPU.

* there's a simple interaction test that shows we do pass the correct data
* there's a slow "CPUDA" test that checks that the device-style update function works, that will run on CI but not generally
* there's a compilation test that runs in the docker container, as for dust

Depends on https://github.com/mrc-ide/odin.dust and https://github.com/mrc-ide/odin/pull/223 (though mostly in order to generate _good_ gpu code)

Fixes #237


docker/Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN install2.r --error \
         remotes \
         socialmixr
 
-RUN Rscript -e 'remotes::install_github(c("mrc-ide/dust",  "mrc-ide/mcstate",  "mrc-ide/eigen1"))'
+RUN Rscript -e 'remotes::install_github(c("mrc-ide/dust", "mrc-ide/mcstate", "mrc-ide/eigen1", "mrc-ide/odin@mrc-2252", "mrc-ide/odin.dust@i75-scope-2"))'
```
johnlees, line 24: Can unpin these versions now?

docker/compile_cuda_model
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env Rscript
+message("Checking CUDA configuration")
+dust::dust_cuda_configuration(quiet = FALSE)
+message("Compiling...")
+gen <- sircovid::compile_gpu(gpu = TRUE, verbose = TRUE)
+message("Initialising...")
+p <- sircovid::carehomes_parameters(
+  sircovid::sircovid_date("2020-02-07"), "england")
+mod <- gen$new(p, 0, 10, seed = 1L)
+message("Running...") # just on the CPU of course
```
johnlees, line 10:
```suggestion
message("Running (CPU)...") # just on the CPU of course
```

tests/testthat/test-gpu.R
```diff
@@ -0,0 +1,60 @@
...
+  expect_identical(
+    compile_gpu(),
+    carehomes)
+  mockery::expect_called(mock_odin_dust, 1L)
+  expect_equal(
+    mockery::mock_args(mock_odin_dust)[[1]],
+    list(sircovid_file("odin/carehomes.R"), real_t = "float", gpu = TRUE))
+
+  compile_gpu(real_t = "double", rewrite_dims = TRUE,
+              gpu = FALSE, gpu_generate = TRUE)
```
johnlees, line 18: Does it make sense to try some of the other CUDA options here? e.g. fast_math, profile

johnlees (changes requested):

docker/Dockerfile
richfitz: yes, good catch

tests/testthat/test-gpu.R
richfitz: I don't think so - this sort of testing should be done in dust (where we don't suffer 1 minute compile times!), and is of relatively limited value unless we check that they had an effect. We do an interaction test already which confirms that the options would be passed

johnlees (approved):

Issue #245, opened 2021-03-15
immune escape features


annecori:
- [x] differential efficacy of vaccine (all 4 rel_* parameters) by variant
- [x] allow infection by one variant before losing natural immunity from the other --> suggest doing this by using "variants" 3 and 4 to model infection with v1 then v2 and v2 then v1 respectively. 
- [x] different duration of infectious period by variant
- [x] different clinical progression by variant
- [x] rate of importation of new variant
- [x] cross-immunity

annecori:
@annecori  working on bullet point 1
@RaphaelS1 working on bullet points 3-5

RaphaelS1:
Point 5 in https://github.com/mrc-ide/sircovid/pull/252
Will start 3 next

RaphaelS1:
Point 3 in https://github.com/mrc-ide/sircovid/pull/256/
Will start 4 next

RaphaelS1:
Point 4 in https://github.com/mrc-ide/sircovid/pull/261 - Needs tests
Starting 2

RaphaelS1:
PR https://github.com/mrc-ide/sircovid/pull/261 is finished, all tests passing and ready to review.
Point 2 in https://github.com/mrc-ide/sircovid/pull/262

RaphaelS1:
Final point in https://github.com/mrc-ide/sircovid/pull/273

RaphaelS1:
Shall we close this @annecori  ?

PR #256, opened 2021-03-22
Add strain_rel_gamma for varying gammas by strains

RaphaelS1:
* Add `strain_rel_gamma` to `carehomes_parameters` for varying gamma_A,gamma_P,gamma_C_1,gamma_C_2 by strain relative to set default
* https://github.com/mrc-ide/sircovid/issues/245 bullet point 3


tests/testthat/test-carehomes-multistrain.R
```diff
@@ -764,3 +764,241 @@ test_that("calculate Rt with both second variant and vaccination", {
...
 })
+
+test_that("strain_rel_gamma works as expected in carehomes_parameters", {
+  expect_silent(carehomes_parameters(sircovid_date("2020-02-07"), "england",
+                                     strain_rel_gamma_A = 1,
+                                     strain_rel_gamma_P = 1,
+                                     strain_rel_gamma_C_1 = 1,
+                                     strain_rel_gamma_C_2 = 1))
+  expect_silent(carehomes_parameters(sircovid_date("2020-02-07"), "england",
+                                     strain_rel_gamma_A = 1,
```
annecori, line 775: you are going to need to change those tests if you follow my suggestion of forcing the first value in each of those to be 1 (so that the 1st strain is the reference)

annecori (changes requested):
This looks good; my only suggestion is to force the first value of all of the relative gammas to be 1; this way the gamma values correspond to strain 1, which is the reference, and the relative gamma parameters are only used to set different values for strain 2 onwards relative to strain 1. This is the approach we have used at the moment for strain_transmission I think and for all the parameters related to vaccine efficacy (rel_susceptbibility for example). This will require more checks (that the 1st value of each is 1 and subsequent values are >=0) and adaptation of current tests which don't always conform to this format.

R/carehomes.R
```diff
@@ -75,6 +75,26 @@ NULL
...
+##' @param strain_rel_gamma_C_1 Vector of relative rates of gamma_C_1 for each
+##'   strain modelled. If `1` all strains have same rates. Otherwise vector of
+##'   same length as `strain_transmission`, with entries that determines the
+##'   relative scaling of the defaults for each strain.
+##'
+##' @param strain_rel_gamma_C_2 Vector of relative rates of gamma_C_2 for each
+##'   strain modelled. If `1` all strains have same rates. Otherwise vector of
+##'   same length as `strain_transmission`, with entries that determines the
+##'   relative scaling of the defaults for each strain.
+##'
```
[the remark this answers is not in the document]
RaphaelS1: yes I had thought similar

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -764,3 +760,248 @@ test_that("calculate Rt with both second variant and vaccination", {
...
+  p <- carehomes_parameters(sircovid_date("2020-02-07"), "england",
+                            strain_transmission = c(1, 1),
+                            strain_rel_gamma_A = c(1, 1.5),
+                            strain_rel_gamma_P = c(1, 1.5),
+                            strain_rel_gamma_C_1 = c(1, 1.5),
+                            strain_rel_gamma_C_2 = c(1, 1.5),
+                            strain_seed_date =
+                              rep(sircovid_date("2020-02-07"), 2),
+                            strain_seed_rate = c(10, 0))
+  p$gamma_A[1] <- p$gamma_A[1] * 1.5
```
annecori, line 854: I don't think you need to do this - even if gamma_A is unchanged for strain 1 but is 1.5 times higher for strain 2 then the overall reproduction number should be lower

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -764,3 +760,248 @@ test_that("calculate Rt with both second variant and vaccination", {
...
+  p <- carehomes_parameters(sircovid_date("2020-02-07"), "england",
+                            strain_transmission = c(1, 1),
+                            strain_rel_gamma_A = 1,
+                            strain_rel_gamma_P = 1,
+                            strain_rel_gamma_C_1 = 1,
+                            strain_rel_gamma_C_2 = 1,
+                            strain_seed_date =
+                              rep(sircovid_date("2020-02-07"), 2),
+                            strain_seed_rate = c(10, 0))
+  p$gamma_P[] <- 0
```
annecori, line 925: I would maybe add a test where you set strain_rel_gamma_P <- c(1, 0) and then check that you still have progression for strain 1 but not for strain 2

annecori (changes requested):
couple extra comments

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -1012,50 +1012,100 @@ test_that("Stuck when gamma =  0", {
...
-                            strain_rel_gamma_C_1 = c(1, 1),
-                            strain_rel_gamma_C_2 = c(1, 1),
+                            strain_rel_gamma_C_1 = 1,
+                            strain_rel_gamma_C_2 = 1,
                             strain_seed_date =
-                              c(sircovid_date("2020-02-07"),
-                                sircovid_date("2020-02-08")),
+                              rep(sircovid_date("2020-02-07"), 2),
                             strain_seed_rate = c(10, 0))
+  p$gamma_P = c(0, 1)
```
annecori: I don't understand why you need this? maybe I'm misunderstanding sth

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -1012,50 +1012,100 @@ test_that("Stuck when gamma =  0", {
...
   end <- sircovid_date("2020-05-01") / p$dt
   steps <- seq(initial$step, end, by = 1 / p$dt)
 
   set.seed(1)
   y <- mod$simulate(steps)
-  expect_equal(p$gamma_P[2], 0)
-  expect_false(all(unlist(y[index_I_C_1_strain_1, , ]) == 0))
-  expect_true(all(unlist(y[index_I_C_1_strain_2, , ]) == 0))
-  expect_false(all(unlist(y[index_I_P_strain_2, , ]) == 0))
+  expect_true(all(unlist(y[index_I_C_1, , ]) == 0))
```
annecori, line 1038: I think the previous test was better because it was able to test that you could stop progression for one strain but not the other

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -1012,50 +1012,100 @@ test_that("Stuck when gamma =  0", {
...
+  p <- carehomes_parameters(sircovid_date("2020-02-07"), "england",
+                            strain_transmission = c(1, 1),
+                            strain_rel_gamma_A = 1,
+                            strain_rel_gamma_P = 1,
+                            strain_rel_gamma_C_1 = 1,
+                            strain_rel_gamma_C_2 = 1,
+                            strain_seed_date =
+                              rep(sircovid_date("2020-02-07"), 2),
+                            strain_seed_rate = c(10, 0))
+  p$gamma_C_1[] <- 0
```
annecori, line 1051: for this part of the test I would also implement something where you can check that you are able to stop progression for strain 2 while keeping progression for strain 1

R/carehomes.R
```diff
@@ -366,12 +366,16 @@ carehomes_parameters <- function(start_date, region,
   severity$psi_star <- severity$p_star / max(severity$p_star)
   severity$p_star_step <- max(severity$p_star)
 
-  strain_rel_gamma_A <- recycle(strain_rel_gamma_A, length(strain_transmission))
-  strain_rel_gamma_P <- recycle(strain_rel_gamma_P, length(strain_transmission))
-  strain_rel_gamma_C_1 <- recycle(strain_rel_gamma_C_1,
-                                  length(strain_transmission))
-  strain_rel_gamma_C_2 <- recycle(strain_rel_gamma_C_2,
-                                  length(strain_transmission))
+  strain_rel_gamma_A <- mcstate:::recycle(assert_relatives(strain_rel_gamma_A),
```
annecori, line 378: for all of these can you please also add an assert non negative statement please? thx

annecori: never mind seems to be there already - ignore me

PR #261, opened 2021-03-24
Variant varying progression

RaphaelS1:
* Updates P_G_D, P_H_D, P_W_D, P_ICU_D to be variant varying


NEWS.md
```diff
@@ -1,3 +1,7 @@
+# sircovid 0.10.22
```
annecori, line 1: can you add news for this version please? and also bump the package version number

R/carehomes.R
```diff
@@ -75,6 +75,33 @@ NULL
...
+##'   strain modelled. If `1` all strains have same rates. Otherwise vector of
+##'   same length as `strain_transmission`, with entries that determines the
+##'   relative scaling of the defaults for each strain.
+##'
+##' @param strain_rel_gamma_C_2 Vector of relative rates of gamma_C_2 for each
+##'   strain modelled. If `1` all strains have same rates. Otherwise vector of
+##'   same length as `strain_transmission`, with entries that determines the
+##'   relative scaling of the defaults for each strain.
+##'
+##' @param strain_rel_severity Vector of relative severity for each
```
annecori, line 98: can we rename this strain_rel_p_D please (severity can be measured in many ways and here we are really only modifying the probability of death so will be clearer I think)

annecori: actually seeing the rest of the code maybe this is fine but then in the documentation can you say it only is applied to probability of death please

R/carehomes.R
```diff
@@ -1235,3 +1304,18 @@ carehomes_data <- function(data, start_date, dt) {
...
+
+scale_severity <- function(severity, strain_rel_severity,
+                           which = c("p_G_D", "p_H_D", "p_W_D", "p_ICU_D")) {
+  severity[which] <- lapply(severity[which], function(x) {
+    x <- matrix(x, nrow = length(x), ncol = length(strain_rel_severity))
+    prob <- matrix(strain_rel_severity, nrow = nrow(x),
+                  ncol = length(strain_rel_severity), byrow = TRUE)
+    x <- x * prob
+    x[x > 1] <- 1
+    x[x < 0] <- 0
```
annecori, line 1317: this really should not happen - probably best to (or in addition to this) check that strain_rel_severity only has non-negative entries. As before could you please also make sure the first value is always 1 so strain 1 is the reference

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -764,3 +764,380 @@ test_that("calculate Rt with both second variant and vaccination", {
...
+
+
+test_that("G_D empty when p_G_D = 0", {
+  np <- 3L
+  p <- carehomes_parameters(sircovid_date("2020-02-07"), "england",
+                            strain_transmission = c(1, 1),
+                            strain_seed_rate = c(1, 0),
+                            strain_seed_date =
+                              sircovid_date(c("2020-02-07", "2020-02-08")))
+  p$p_G_D[] <- 0
```
annecori, line 1077: with the same approach as for the gammas you could check that if strain_rel_severity = c(1, 0) then you get no-one in G_D for strain 2 but still some people in G_D_ for strain 1

annecori (commented):
will review again especially the Rt calculation once these have been addressed and PR #256 has been merged onto master so it's easier to see the specific changes associated with the current PR

NEWS.md
RaphaelS1: I'm leaving this for now as it keeps conflicting with my other branches. Will resolve after https://github.com/mrc-ide/sircovid/pull/256 is merged

tests/testthat/test-carehomes-check.R
```diff
@@ -953,3 +953,24 @@ test_that("Individuals can infect in compartment with non-zero transmission", {
   helper("ICU_transmission", "ICU_W_R_unconf", "gamma_ICU_W_R_step")
   helper("ICU_transmission", "ICU_W_R_conf", "gamma_ICU_W_R_step")
 })
+
+
+test_that("No one is hospitalised, no-one recovers in edge case", {
```
annecori, line 958: same comment as in the previous PR i.e. could we also add a test which checks this when setting the probabilities of death to zero ONLY for strain 2 using the strain_rel_severity parameter?

annecori:
@RaphaelS1  please check you are happy with the tests I added and bug I fixed? 
then if you are happy and if all checks are passing we can merge

annecori (approved):

PR #262, opened 2021-03-25
Immune escape multivoc

RaphaelS1:
See https://github.com/mrc-ide/sircovid/issues/245

Ref:

VOC 3 = VOC 1 -> VOC 2
VOC 4 = VOC 2 -> VOC 1

To-do:

* [x] 1. Force `n_strains` to either 1 (one VOC) or 2 (two VOC + two interactions)
```r
  if (length(strain_transmission) > 2) {
    stop("Only 1 or 2 strains valid ('strain_transmission' too long)'.")
  }
```
* [x] 2. Add 'variant 3' = variant 1.2 (R_1 -> E_3 (E_12)) and 'variant 4' = variant 2.1 (R_2 -> E_4 (E_21))
```r
n_strains_user <- user()
n_strains <- if (n_strains_user == 2) 4 else 1
```
* [x] 3. Set `p_SE3 = 0` and `p_SE4 = 0` and `p_E2 = 1 - p_E1` 
```r
n_S_progress[4, 3:4, 1] <- 0
```
* [x] 4. Add `p_R2E4 >= 0` and `p_R1E3 >= 0` s.t. `p_R1S = 1 - p_R1E3` and `p_R2S = 1 - p_R2E4`
* [x] 5. `p_R1E2 = p_R1E4 = p_R2E3 = p_R2E1 = 0`
* [x] 6. Transitions from `E_3/4` -> `I_3/4` -> `R_3/4` as normal
* [x] 7. p_R1E3 depends on vax status, vax efficacy against VOC2
* [x] 8. p_R2E4 depends on vax status, vax efficacy against VOC1
* [ ] 9. Update Rt calculations
* [ ] 10. Old tests passing
* [ ] 11. New tests passing

image


inst/odin/carehomes.R
```diff
@@ -42,7 +42,9 @@ n_I_P_vaccinated[, ] <-
   sum(n_II_P_next_vacc_class[i, , , j])
 dim(n_I_P_vaccinated) <- c(n_groups, n_vacc_classes)
 n_R_vaccinated[, ] <-
-  sum(n_R_next_vacc_class[i, , j]) + sum(n_RS_next_vacc_class[i, , j])
+  sum(n_RR_next_vacc_class[i, , j]) +
+  sum(n_RS_next_vacc_class[i, , j]) +
+  sum(n_RE_next_vacc_class[i, , j])
```
annecori, line 47: not sure how you have defined those but looks like the n_RE should come from another strain so should perhaps be indexed differently? Please check

inst/odin/carehomes.R
```diff
@@ -356,37 +363,81 @@ n_II_P[, , , ] <- n_I_P_progress[i, j, k, l] -
...
+  (n_R_progress[i, j, k])))
+
+## Numbers going from R to E
+##   R -> E1 and E2 are always zero
+##   nR1E3 are Number leaving R1 and not going to S;
+##     nR2E3 eq nR3E3 eq n_R4E3 eq 0;
+##   nR2E4 is Number leaving R2 and not going to S;
+##     nR2E4 gt 0; nR1E4 eq n_R3E4 eq n_R4E4 eq 0;
+## Total possible to E3 and E4 are from R1 and R2 respectively
+n_RE[, , ] <- if (j < 3) 0 else
```
annecori, line 416: would be good to clarify how n_RE is indexed, it looks like n_RE[i, j, k] is the number of individuals moving from R[i,j-2,k] to E[I,j,k] so that it is indexed by the recipient E compartment (although I may be misunderstanding this).

inst/odin/carehomes.R
```diff
@@ -264,11 +265,13 @@ gamma_ICU_pre <- if (as.integer(step) >= length(gamma_ICU_pre_step))
...
 n_S_progress[, , ] <- if (j == 1)
   rbinom(n_S_progress_tot[i, k], lambda[i, 1] / sum(lambda[i, ])) else
-    rbinom(n_S_progress_tot[i, k] - sum(n_S_progress[i, 1:(j - 1), k]),
-           lambda[i, j] / sum(lambda[i, j:n_strains]))
+    (if (j == 2) n_S_progress_tot[i, k] - n_S_progress[i, 1, k] else 0)
 
 ## Introduction of new strains. n_S_progress is arranged as:
 ##
+## FIXME (RS): Is the below correct? Should it actually say
+##   [age, strain, stage]?
```
annecori, line 273: yes you are right the comment is wrong

inst/odin/carehomes.R
```diff
@@ -1460,7 +1527,9 @@ update(react_pos) <- sum(new_T_PCR_pos[2:18, , , ])
 
 ## prob_strain gives probability of an infection in group i being of strain j
 initial(prob_strain[, ]) <- 0
-dim(prob_strain) <- c(n_groups, n_strains)
+dim(prob_strain) <- c(n_groups, n_real_strains)
```
annecori, line 1530: I think given what we have decided to do for the Rt calculation we could drop the n_group dimension. I've tried to do this in a later commit but didn't go very far

inst/odin/carehomes.R
```diff
@@ -42,7 +42,9 @@ n_I_P_vaccinated[, ] <-
   sum(n_II_P_next_vacc_class[i, , , j])
```
annecori, line 42: could you add some comments at the start of the odin code to explain how strains 3 and 4 are used i.e. which corresponds to which?

RaphaelS1:
See #269

PR #266, opened 2021-03-26
I265 catchup

annecori:


lwhittles (approved):
Checked using full pipeline from fitting through to vaccine simulation - all working and producing sensible results

PR #267, opened 2021-03-30
Collect age/vaccination disaggregated deaths/admission

richfitz:
Also includes state upgrade function


lwhittles (approved):
looks great, checked with rtm_inference_simulation and all runs fine

PR #269, opened 2021-04-01
Immune escape v2

annecori:
builds on [https://github.com/mrc-ide/sircovid/pull/262](https://github.com/mrc-ide/sircovid/pull/262)

Current issues that need dealing with: 

- [x] Check whether we need the R (recovered) in the reproduction number calculation - can we ignore and set to NULL for single strain - or is that already the case? 

Series of failing tests:

- [x] the usual number not exactly matching in here:
test_file("./tests/testthat/test-carehomes-support.R")

- [x] ## annotated but not totally fixed:
test_file("./tests/testthat/test-carehomes-multistrain.R")
test_file("./tests/testthat/test-carehomes-rt.R")
test_file("./tests/testthat/test-carehomes-pmcmc.R") 

- [x] ## this requires fixing future beta - may require help from @lwhittles and @richfitz  who wrote those functions in the first place
test_file("./tests/testthat/test-support.R")


R/carehomes.R
```diff
@@ -186,10 +187,18 @@ NULL
 ##'   rate is used for all age groups; if a vector of values if used it should
 ##'   have one value per age group.
 ##'
-##' @param model_pcr_and_serology_user A value of 1 or 0 so switch on or off the
+##' @param model_pcr_and_serology A value of 1 or 0 so switch on or off the
```
richfitz, line 190: can you let Ed know you changed that, I think he's changing this too soon

R/utils.R
```diff
@@ -193,3 +193,19 @@ gamma_mucv2shapescale <- function(mu, cv) {
   scale <- mu * cv^2
   list(shape = shape, scale = scale)
 }
+
+
+mirror <- function(x) c(x, rev(x))
```
richfitz, line 198:
```suggestion
mirror <- function(x) {
  c(x, rev(x))
}
```

richfitz: I also wonder if this should be `mirror_strain()` unless this will be useful elsewhere

R/utils-assert.R
```diff
@@ -66,11 +66,19 @@ assert_proportion <- function(x, name = deparse(substitute(x))) {
...
+
   assert_relatives <- function(x, name = deparse(substitute(x))) {
     if (x[[1]] != 1) {
       stop(sprintf("'%s[[1]]' must be 1", name), call. = FALSE)
     }
     assert_non_negative(x, name)
     invisible(x)
  }
+
+  assert_01 <- function(x, name = deparse(substitute(x))) {
```
richfitz, line 78: How about accepting logical too?

richfitz (changes requested):
I can't speak to the epi, but the code looks fine. I am concerned about the possibility of future changes to the rt calculation, but nothing you can easily do about that. A few minor changes throughout

R/utils-assert.R
RaphaelS1: Does odin accept logicals? I guess as.integer works on logicals too

richfitz: yes, that was my thought. you'll still pass 0/1 to odin, but at the R level could accept FALSE/TRUE and run them through as.integer

inst/odin/carehomes.R
```diff
@@ -1458,10 +1545,18 @@ initial(react_pos) <- 0
...
-## prob_strain gives probability of an infection in group i being of strain j
-initial(prob_strain[, ]) <- 0
-dim(prob_strain) <- c(n_groups, n_strains)
-update(prob_strain[, ]) <- lambda[i, j] / sum(lambda[i, ])
+## grp_prob_strain is probability of an infection in group i being of strain j
+## prob_strain is probability of an infection being of strain j
+grp_prob_strain[, ] <- lambda[i, j] / sum(lambda[i, ])
+dim(grp_prob_strain) <- c(n_groups, n_real_strains)
+
+tmp_prob_strain <- sum(lambda[, 1]) / sum(lambda[, ])
```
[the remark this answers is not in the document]
RaphaelS1: No it's called tmp because it's used once in line below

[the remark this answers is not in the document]
RaphaelS1: Updated in https://github.com/mrc-ide/sircovid/pull/272

R/utils-assert.R
RaphaelS1: Updated in https://github.com/mrc-ide/sircovid/pull/272

inst/odin/carehomes.R
```diff
@@ -456,16 +516,31 @@ n_EI_P_next_vacc_class[, , ] <- n_EE_next_vacc_class[i, j, k_E, k] -
...
   n_EE[i, j, k, l] -
   n_EE_next_vacc_class[i, j, k, l] -
   n_E_next_vacc_class[i, j, k, l] +
-  (if (l == 1) n_E_next_vacc_class[i, j, k, n_vacc_classes] else
-    n_E_next_vacc_class[i, j, k, l - 1]) +
-  (if (k == 1) (if (l == 1) n_SE_next_vacc_class[i, j, n_vacc_classes] else
-    n_SE_next_vacc_class[i, j, l - 1]) else
-      (if (l == 1) n_EE_next_vacc_class[i, j, k - 1, n_vacc_classes] else
-        n_EE_next_vacc_class[i, j, k - 1, l - 1]))
+  (if (l == 1)
```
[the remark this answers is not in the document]
RaphaelS1: Not sure from my side

R/utils-assert.R
```diff
@@ -66,11 +66,19 @@ assert_proportion <- function(x, name = deparse(substitute(x))) {
   }
   invisible(x)
 }
-  
+
```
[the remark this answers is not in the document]
RaphaelS1: Updated in #272

inst/odin/carehomes.R
```diff
@@ -42,7 +44,9 @@ n_I_P_vaccinated[, ] <-
   sum(n_II_P_next_vacc_class[i, , , j])
 dim(n_I_P_vaccinated) <- c(n_groups, n_vacc_classes)
 n_R_vaccinated[, ] <-
-  sum(n_R_next_vacc_class[i, , j]) + sum(n_RS_next_vacc_class[i, , j])
+  sum(n_RR_next_vacc_class[i, , j]) +
```
annecori, line 47: I would perhaps rename n_RR_next_vacc_class to n_R_next_vacc_class as they move vaccination class but they don't move from R to the next R compartment which the current name suggests I think.

inst/odin/carehomes.R
```diff
@@ -356,37 +363,84 @@ n_II_P[, , , ] <- n_I_P_progress[i, j, k, l] -
...
+
+## 1) R -> S vaccine progression
+n_RS_next_vacc_class[, , ] <- if (n_vacc_classes == 1) 0 else
+  rbinom(n_RS[i, j, k], p_R_next_vacc_class[i, j, k])
+n_RS_same_vacc_class[, , ] <- n_RS[i, j, k] - n_RS_next_vacc_class[i, j, k]
+## 2) R -> E vaccine progression
+n_RE_next_vacc_class[, , ] <- if (n_vacc_classes == 1) 0 else
+  rbinom(n_RE[i, j, k], p_R_next_vacc_class[i, j, k])
+n_RE_same_vacc_class[, , ] <- n_RE[i, j, k] - n_RE_next_vacc_class[i, j, k]
+## 3) R -> R vaccine progression
```
annecori, line 431: for all that section 3) R->R I would probably keep the code as it was: I don't think you need `n_RR` at all as we don't have sequential R compartments. So we just need to know how many people in R get vaccinated and stay in R, and that's exactly what we had before. So I don't think we need the extra variables `n_RR` and `n_RR_same_vacc_class`, and I think we can keep the simpler (and I think less misleading naming) of `n_R_next_vacc_class` rather than `n_RR_next_vacc_class`

R/carehomes.R
```diff
@@ -186,10 +187,18 @@ NULL
...
 ##'   flows out of T_PCR_neg and T_sero_neg and the corresponding cap on the
 ##'   number of individuals leaving the R compartments
 ##'
+##' @param model_super_infection A value of 1 or 0 so switch on or off
```
annecori, line 194: could do this later but would be good to add some examples showing how to set parameters for the multi strain model

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -453,49 +610,162 @@ test_that("Cannot calculate Rt for multistrain without correct inputs", {
...
+              "Expect elements of Rt to have dimensions")
 })
 
 
 test_that("Can calculate Rt with an empty second variant ", {
   ## Run model with 2 variants, but both have same transmissibility
   ## no seeding for second variant so noone infected with that one
   p <- carehomes_parameters(sircovid_date("2020-02-07"), "england",
-                            strain_transmission = c(1, 1))
+                            strain_transmission = c(1, 0),
```
annecori, line 759: I think this slightly changes the nature of the test - why did you have to set the second element of strain_transmission to zero? should still work with 1?

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -517,38 +790,62 @@ test_that("Can calculate Rt with an empty second variant ", {
...
 test_that("Can calculate Rt with a second less infectious variant", {
   ## seed with 10 cases on same day as other variant
-
   p <- carehomes_parameters(sircovid_date("2020-02-07"), "england",
                             strain_transmission = c(1, 0.1),
                             strain_seed_date =
                               sircovid_date(c("2020-02-07", "2020-02-08")),
-                            strain_seed_rate = c(10, 0))
+                            strain_seed_rate = c(10, 0),
+                            strain_rel_severity = c(1, 0.8),
```
annecori, line 839: I would keep the severity for another test and keep it to the default c(1,1) in this one

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -570,25 +870,26 @@ test_that("Can calculate Rt with a second less infectious variant", {
...
+  mod$set_index(c(index_S, index_R))
   y <- mod$simulate(steps)
 
-  rt_1_single_class <- carehomes_Rt(steps, y[, 1, ], p)
-  rt_all_single_class <- carehomes_Rt_trajectories(steps, y, p)
+  S <- y[1:19, , ]
+  R <- y[20:38, , ]
+
+  rt_1_single_class <- carehomes_Rt(steps, S[, 1, ], p, R = R[, 1, ])
+  rt_all_single_class <- carehomes_Rt_trajectories(steps, S, p, R = R)
```
annecori, line 888: I would rescue the comment you deleted here

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -1134,13 +1493,15 @@ test_that("Lower rate variant has higher Rt", {
...
   prob_strain <- y[index_prob_strain, , ]
 
-  rt_1_all <- carehomes_Rt_trajectories(steps, S, p, prob_strain)
+  rt_1_all <- carehomes_Rt_trajectories(steps, S, p, prob_strain, R = R,
+                                        weight_Rt = TRUE)
 
   ## Rt should be higher (or equal) for the two variant version
-  expect_true(all(rt_1_all$Rt_all <= rt_15_all$Rt_all))
-  expect_true(all(rt_1_all$Rt_general <= rt_15_all$Rt_general))
+  expect_true(all(rt_1_all$Rt_all - 0.001 <= rt_15_all$Rt_all))
```
annecori, line 1503: could perhaps replace with your nicer function

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -1476,32 +1850,148 @@ test_that("G_D empty when p_G_D = 0", {
...
+  set.seed(1)
+  y <- mod$transform_variables(
+    drop(mod$simulate(steps)))
+  expect_true(any(y$E[, 1, , , , 2] > 0))
+  expect_true(any(y$E[, 2, , , , 2] > 0))
+  expect_true(all(y$E[, 3:4, , , , 2] == 0))
+})
+
+
+test_that("Everyone in R1 when strain_transmission = c(1, 0)", {
```
annecori, line 1908: I would rename "Nobody in R2-R4" rather than "Everyone in R1"

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -1476,32 +1850,148 @@ test_that("G_D empty when p_G_D = 0", {
...
+  y <- mod$transform_variables(
+    drop(mod$simulate(steps)))
+
+  expect_true(all(y$S[, , , 85] == 0))
+  expect_true(all(y$R[, 2:4, , , 85] == 0))
+  expect_false(all(y$R[, 1, , , 85] == 0))
+})
+
+
+test_that("Can only move from R3 and R4 to S", {
```
annecori, line 1934: I would rename the test "Can only move to S from R3 and R4" ( you can move to S from R1 and R2)

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -1476,32 +1850,148 @@ test_that("G_D empty when p_G_D = 0", {
...
+    drop(mod$simulate(steps)))
 
-  # Strain 1
-  expect_false(all(y[index_G_D_strain_1, , ] == 0))
-  expect_false(all(y[index_G_D_strain_1, , ] == 0))
-  # Strain 2
-  expect_true(all(y[index_G_D_strain_2, , ] == 0))
-  expect_true(all(y[index_G_D_strain_2, , ] == 0))
+  expect_true(all(y$R[, 1:2, , 450] == 0))
+  expect_false(all(y$R[, 3:4, , 450] == 0))
```
annecori, line 1996: I was going to suggest you could do better than this and check that the sum of individuals in R[3:4] at last time step is equal to the total population (p$N_tot), or not far off. But when I checked this locally, turns out there is almost noone in R[4], and that remains true even when I set strain_seed_rate to something crazy high. So I am not sure the seeding works properly... This needs checking

tests/testthat/test-carehomes-pmcmc.R
```diff
@@ -214,7 +214,7 @@ test_that("can combine rt calculations over trajectories without reordering", {
 test_that("adding incidence adds appropriate states - nested", {
   dat <- reference_data_mcmc()
   dat$trajectories$state <- array(
-    dat$trajectories$state, c(129, 11, 2, 32),
+    dat$trajectories$state, c(111, 11, 2, 32),
```
annecori, line 217: why has this changed? just curious?

RaphaelS1: I believe because prob strain went from along groups (19) to not (1) so 18 difference

tests/testthat/test-carehomes-multistrain.R
RaphaelS1: Found the same, which is why the values are set so high

tests/testthat/test-carehomes-multistrain.R
annecori: "## Rt should be lower (or equal) for the two variant version"

richfitz (approved):

R/carehomes_rt.R
```diff
@@ -347,69 +463,52 @@ carehomes_Rt_mean_duration_weighted_by_infectivity <- function(step, pars,
...
   ## Account for different infectivity levels depending on vaccination stage
-
   mean_duration <- mean_duration *
     outer(outer(pars$rel_infectivity, rep(1, n_time_steps)), rep(1, n_strains))
 
+  ## Account for different infectivity levels depending on strain
+  mean_duration <-
+    aperm(aperm(mean_duration, c(4, 1, 2, 3)) * pars$strain_transmission,
+          c(2, 3, 4, 1))
+
```
edknock, line 508: Not sure I understand what is going on with the `aperm`s here, so maybe useful to have a bit more commenting?

RaphaelS1: Fixed in https://github.com/mrc-ide/sircovid/pull/272

PR #271, opened 2021-04-08
Relax waning immunity restrictions

edknock:
Removing the `model_pcr_and_serology_user` parameter, and no longer restricting waning immunity by capping with the numbers in `T_sero_neg` and `T_PCR_neg`.

Note we are now no longer removing individuals from `T_sero_neg` and `T_PCR_neg` so `N_tot2` and `N_tot3` can increase when waning immunity is allowed.

These compartments are not used in the likelihood calculation - we just use `N_tot_16_64` and `N_tot_react` as the denominator. One issue is that it is now possible to have more "people" in the seropositive compartment than in the population, i.e. `model_sero_pos > N_tot_16_64` - to deal with this I have capped `model_sero_pos` in the compare with `N_tot_16_64` to avoid a binomial probability > 1. I have also done similarly for the PCR positives in the compare. I would expect we are highly unlikely to encounter this case in practice though - it would require faster waning immunity than waning PCR or sero positivity and high rates of infection.


tests/testthat/test-carehomes-vaccination.R
```diff
@@ -1276,12 +1265,39 @@ test_that("N_tot, N_tot2 and N_tot3 stay constant with vaccination", {
 })
 
 
-test_that("N_tot, N_tot2, N_tot3 are constant with vaccination and k > 1", {
+test_that("N_tot stays constant with vaccination and waning immuity, while
```
annecori, line 1268: typo in "Immunity"

annecori (approved):

PR #272, opened 2021-04-08
Efficacy by strain v2

RaphaelS1:


inst/odin/carehomes.R
```diff
@@ -1511,18 +1510,15 @@ update(react_pos) <- sum(new_T_PCR_pos[2:18, , , ])
 
 ## grp_prob_strain is probability of an infection in group i being of strain j
 ## prob_strain is probability of an infection being of strain j
-grp_prob_strain[, ] <- lambda[i, j] / sum(lambda[i, ])
-dim(grp_prob_strain) <- c(n_groups, n_real_strains)
+grp_prob_strain[, , ] <- lambda_susc[i, j, k] / sum(lambda_susc[i, , k])
```
annecori, line 1513: it would be good to discuss this with @edknock. I am not sure we want to weigh this by the relative susceptibility.

inst/odin/carehomes.R
```diff
@@ -1511,18 +1510,15 @@ update(react_pos) <- sum(new_T_PCR_pos[2:18, , , ])
...
+dim(grp_prob_strain) <- c(n_groups, n_real_strains, n_vacc_classes)
 
-tmp_prob_strain <- sum(lambda[, 1]) / sum(lambda[, ])
+prob_strain_1 <- sum(lambda_susc[, 1, ]) / sum(lambda_susc[, , ])
```
annecori, line 1516: same here @edknock

inst/odin/carehomes.R
annecori: actually I'm pretty sure this should be using lambda not lambda_susc as it ignores how many susceptible are in each vaccination class

annecori (changes requested):
This looks fine to be except I think that prob_strain should be calculated as before using lambda not lambda_susc

edknock (approved):

annecori (approved):

PR #279, opened 2021-04-12
Cleanup assertions and vectorised expectations

### LINKED ISSUE #274: Cleanup assertions and expect_ functions
Currently some assertions only work for scalars (e.g. have `&&` or `||` in them) whereas others are vectorised; it would probably be cleanest to either vectorise all of them (as we have `assert_scalar` when needed) or to have a set of scalars and a set for vectors.

Similarly, I've started adding a few vectorised `expect_` tests because most of our tests are things like `expect_true(all(x == y))` but it would be cleaner to have a vectorised `expect_equal`, e.g. `expect_v_equal(x, 0)` which internally calls the same thing; I've already done this for lt, gt, lte, gte, and included rounding and tolerance levels as that tends to be pretty common too

Fairly low priority though...

RaphaelS1:
Closes #274

### DIFF
--- a/R/utils-assert.R
+++ b/R/utils-assert.R
@@ -67,6 +67,7 @@ assert_proportion <- function(x, name = deparse(substitute(x))) {
   invisible(x)
 }
 
+
 assert_relatives <- function(x, name = deparse(substitute(x))) {
   if (x[[1]] != 1) {
     stop(sprintf("'%s[[1]]' must be 1", name), call. = FALSE)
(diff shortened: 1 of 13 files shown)


R/utils-assert.R
```diff
@@ -77,8 +78,8 @@ assert_relatives <- function(x, name = deparse(substitute(x))) {
 
 
 assert_logical <- function(x, name = deparse(substitute(x))) {
-  if (!(x %in% 0:1 || is.logical(x))) {
-    stop(sprintf("'%s' must be logical or in {0, 1}", name), call. = FALSE)
+  if (!all(x %in% 0:1 | is.logical(x))) {
```
richfitz, line 81: is.logical applies to whole vector so should use `||`; how about:

```suggestion
  if (!(all(x %in% 0:1) || is.logical(x))) {
```

richfitz (changes requested):
small change here

richfitz (approved):

PR #284, opened 2021-04-15
Improve inputting time-varying parameters

edknock:
The `carehomes_parameters_severity` and `carehomes_parameters_progression` now can be used to create time-varying severity and progression parameters that can be fed into the `carehomes_parameters` function.

Structurally now we have for all gammas for the main pathway durations
- `gamma_X_step` (a vector over steps)
- `rel_gamma_X` (a vector over strains)
(This is for: `gamma_E`, `gamma_A`, `gamma_P`, `gamma_C_1`, `gamma_C_2`, `gamma_ICU_pre`, `gamma_ICU_D`, `gamma_ICU_W_D`, `gamma_ICU_W_R`, `gamma_H_D`, `gamma_H_R`, `gamma_W_D`, `gamma_W_R`, `gamma_G_D`.)

For the main pathway probabilities we have
- `p_X_step` (an array with dimensions steps x groups)
- `rel_p_X` (an array with dimensions groups x strains x vaccine classes)
(This is for `p_C`, `p_H`, `p_ICU`, `p_ICU_D`, `p_H_D`, `p_W_D`, `p_G_D`, `p_star`. Note that instead of `rel_p_C` we have `rel_p_sympt` and instead of `rel_p_H` we have `rel_p_hosp_if_sympt`.)

`sircovid_parameters_beta` has been renamed to `sircovid_parameters_piecewise_linear` and `sircovid_parameters_beta_expand` has been renamed to `sircovid_parameters_expand_step`.


R/carehomes.R
```diff
@@ -817,10 +796,207 @@ carehomes_severity <- function(p) {
...
+##'   hospital). If `value` is `NULL` then the value
+##'   from `severity` is used. If `value` is scalar then `date` can be `NULL` or
+##'   unspecified. If `value` is a vector then `date` must be a vector of
+##'   sircovid dates of the same length as `value`.
+##'
+##' @return A list of severity parameters
+##'
+##' @export
+carehomes_parameters_severity <- function(dt,
+                                          severity = NULL,
```
richfitz, line 880: Generally, I think you could write this with defaults for `p_C` etc as `NULL` which will make this easier to scan.

R/carehomes.R
```diff
@@ -817,10 +796,207 @@ carehomes_severity <- function(p) {
...
+
   severity <- sircovid_parameters_severity(severity)
   severity <- lapply(severity, carehomes_severity)
-  severity$p_G_D[length(severity$p_G_D)] <- p_death_carehome
+
+  get_p_step <- function(x, name) {
+
+    p_name <- paste0("p_", name)
+    p <- x[[p_name]]
+    p_value <- get(p_name)$value
```
richfitz, line 909: Use of `get` is usually a flag you're doing something odd. The trick here is to assemble a list as quickly as possible.

R/carehomes.R
```diff
@@ -817,10 +796,207 @@ carehomes_severity <- function(p) {
...
+
+    x[[paste0(p_name, "_step")]] <- p_step
+    x[[paste0(p_name)]] <- NULL
+    x[[paste0("n_", p_name, "_steps")]] <- dim(p_step)[1]
+
+    x
+  }
+
+  ## Set up time-varying severity parameters
+  time_varying_severity <- c("C", "H", "ICU", "H_D", "W_D",
```
richfitz, line 994: So here, we do:

```r
time_varying_severity <- list(C = C, H = H, ..., star = star)
```

and from that point you don't have to do anything with `get`

R/carehomes.R
```diff
@@ -1070,87 +1246,211 @@ carehomes_parameters_waning <- function(waning_rate) {
...
-##'   same length as `strain_transmission`, with entries that determines the
-##'   relative scaling of the defaults for each strain.
-##'
-##' @param rel_gamma_P Vector of relative rates of gamma_P for each
-##'   strain modelled. If `1` all strains have same rates. Otherwise vector of
-##'   same length as `strain_transmission`, with entries that determines the
-##'   relative scaling of the defaults for each strain.
+##' @param dt The step size
+##'
+##' @param gamma_E A list with `date` and `value` for changes in gamma_E. If
```
richfitz, line 1251: Save yourself some pain to update these and describe the format in the details, then just refer to the values.

```r
##' @section Time-varying parameters:
##' Every time varying parameter has the same format, as a list with 
##'   `date` and `value` for changes in the parameter. If
##'   `value` is `NULL` then a default value is used. If `value` is scalar then
##'   `date` can be `NULL` or unspecified. If `value` is a vector then `date`
##'   must be a vector of sircovid dates of the same length as `value` 
##'
##' [....]
##'
##' @param gamma_E Time-varying parameter describing gamma of E (what is this!?). See Details.
```

richfitz (changes requested):
I have concerns over the two big sections with `get()` as there is a simpler way of doing that. A hint of the pattern is included but this might be easiest/fastest to adjust together

R/carehomes.R
edknock: Good idea, thanks, have done this

R/carehomes.R
edknock: These are all now NULL as default (rather than a list of NULLs)

richfitz (approved):

R/carehomes.R
```diff
@@ -1414,17 +1691,3 @@ carehomes_data <- function(data, start_date, dt) {
...
-
-
-scale_severity <- function(severity, strain_rel_severity,
-                           which = c("p_G_D", "p_H_D", "p_W_D", "p_ICU_D")) {
-  severity[which] <- lapply(severity[which], function(x) {
-    x <- matrix(x, nrow = length(x), ncol = length(strain_rel_severity))
-    prob <- matrix(strain_rel_severity, nrow = nrow(x),
-                  ncol = length(strain_rel_severity), byrow = TRUE)
-    x <- x * prob
-    x[x > 1] <- 1
```
RaphaelS1, line 1426: I can see that we don't need a separate function for these now as it's been generalised but do we have a check like this somewhere else to make sure no probs are above 1?

R/carehomes.R
```diff
@@ -1070,87 +1239,195 @@ carehomes_parameters_waning <- function(waning_rate) {
...
 ##'
-##' @param rel_gamma_A Vector of relative rates of gamma_A for each
-##'   strain modelled. If `1` all strains have same rates. Otherwise vector of
-##'   same length as `strain_transmission`, with entries that determines the
-##'   relative scaling of the defaults for each strain.
+##' @section Time-varying parameters:
+##' Every time varying parameter has the same format, which can be `NULL` (in
+##'   which case a default single value is used) or a list with `date` and
+##'   `value` for changes in the parameter. If `value` is scalar then
+##'   `date` can be `NULL` or unspecified. If `value` is a vector then `date`
```
RaphaelS1, line 1248: Unclear to me what an 'unspecified' date, to me 'unspecified' = NULL and 'missing' = not provided.

R/carehomes.R
```diff
@@ -524,6 +455,54 @@ carehomes_parameters <- function(start_date, region,
...
+    ret$rel_gamma_A <- mirror_strain(strain_rel_gamma_A)
+    ret$rel_gamma_P <- mirror_strain(strain_rel_gamma_P)
+    ret$rel_gamma_C_1 <- mirror_strain(strain_rel_gamma_C_1)
+    ret$rel_gamma_C_2 <- mirror_strain(strain_rel_gamma_C_2)
+  } else {
+    ret$rel_gamma_A <- rep(1, strain$n_strains)
+    ret$rel_gamma_P <- rep(1, strain$n_strains)
+    ret$rel_gamma_C_1 <- rep(1, strain$n_strains)
+    ret$rel_gamma_C_2 <- rep(1, strain$n_strains)
+  }
```
RaphaelS1, line 495: I wonder if we could simplify this slightly, something like (I have not tested this locally, also Rich doesn't like `get` so maybe not the best idea)

```suggestion
strain_ones <- rep(1, strain$n_strains)
for (str in c("rel_gamma_A", "rel_gamma_P", "rel_gamma_C_1", "rel_gamma_C_2")) {
  gamma <- get(sprintf("strain_%s", str))
  gamma <- recycle(assert_relatives(gamma), length(strain_transmission))
  if (length(gamma) == 2) {
    ret[[str]] <- mirror_strain(gamma)
  } else {
    ret[[str]] <- strain_ones 
  }
  ret$rel_gamma_E <- strain_ones
}
```

R/carehomes.R
```diff
@@ -524,6 +455,54 @@ carehomes_parameters <- function(start_date, region,
...
+  ret$rel_gamma_ICU_pre <- rep(1, strain$n_strains)
+  ret$rel_gamma_H_D <- rep(1, strain$n_strains)
+  ret$rel_gamma_H_R <- rep(1, strain$n_strains)
+  ret$rel_gamma_ICU_D <- rep(1, strain$n_strains)
+  ret$rel_gamma_ICU_W_D <- rep(1, strain$n_strains)
+  ret$rel_gamma_ICU_W_R <- rep(1, strain$n_strains)
+  ret$rel_gamma_W_D <- rep(1, strain$n_strains)
+  ret$rel_gamma_W_R <- rep(1, strain$n_strains)
+  ret$rel_gamma_G_D <- rep(1, strain$n_strains)
```
RaphaelS1, line 504: Similarly to above
```suggestion
for (gamma in c("rel_gamma_ICU_pre", "rel_gamma_H_D", "rel_gamma_H_R", "rel_gamma_ICU_D", "rel_gamma_ICU_W_D", "rel_gamma_ICU_W_R", "rel_gamma_W_D", "rel_gamma_W_R", "rel_gamma_G_D")) {
 ret[[gamma]] <- strain_ones                          
}
```

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -1809,77 +1779,6 @@ test_that("No one is hospitalised, no-one recovers in edge case 2 - multi", {
...
-      scale_severity(severity, c(0.1, 0.2))[which],
-      function(x) all(x[, 2] == x[, 1] * 2), logical(1)
-    ))
-  )
-  expect_true(all(unique(unlist(
-    scale_severity(severity, 1e10)[which])) %in% c(1, 0)))
-})
-
-
-test_that("carehomes_parameters with rel_severity works as expected", {
```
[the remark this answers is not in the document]
edknock: Removed this because the `scale_severity` function has been removed

R/carehomes.R
edknock: Have changed this to 'missing' wherever I used 'unspecified'

R/carehomes.R
edknock: Have done this but in a way that avoids using `get`

RaphaelS1 (approved):

PR #285, opened 2021-04-16
fix Rt calculation for NA prob strain

### LINKED ISSUE #276: Change handling of prob_strain NA in Rt calculation
Currently Rt calculation returns all NA if any steps in `prob_strain` are NA

https://github.com/mrc-ide/sircovid/blob/128fec1abc479dc0749a55434e49d148cbf1c3aa/R/carehomes_rt.R#L195-L202

Instead we could change this so Rt is only NA_real_ in the same time steps as prob_strain is NA, otherwise compute Rt.

RaphaelS1:
Closes #276

Previously if any steps in prob_strain were NA then all Rt were NA, now Rt only NA in the same steps as prob_strain

### DIFF
--- a/R/carehomes_rt.R
+++ b/R/carehomes_rt.R
@@ -66,8 +66,8 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
   ## deal with the prob_strain NA case at the beginning before any variables
   ##  are modified
   if (!is.null(prob_strain) && any(is.na(prob_strain))) {
-    which_nna <- !vapply(seq(ncol(prob_strain)),
-                        function(i) any(is.na(prob_strain[, i])), logical(1))
+    which_nna <- !vlapply(seq(ncol(prob_strain)),
+                          function(i) any(is.na(prob_strain[, i])))
 
     ## calculate Rt by first ignoring NA
     ret <- carehomes_Rt(step[which_nna], S[, which_nna], p,
--- a/R/carehomes_rt.R
+++ b/R/carehomes_rt.R
@@ -63,6 +63,39 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
                          eigen_method = "power_iteration", R = NULL,
                          weight_Rt = FALSE) {
 
+  ## deal with the prob_strain NA case at the beginning before any variables
+  ##  are modified
+  if (!is.null(prob_strain) && any(is.na(prob_strain))) {
+    which_nna <- !vapply(seq(ncol(prob_strain)),
+                        function(i) any(is.na(prob_strain[, i])), logical(1))
+
+    ## calculate Rt by first ignoring NA
+    ret <- carehomes_Rt(step[which_nna], S[, which_nna], p,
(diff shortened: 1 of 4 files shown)


R/carehomes_rt.R
```diff
@@ -63,6 +63,39 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
                          eigen_method = "power_iteration", R = NULL,
                          weight_Rt = FALSE) {
 
+  ## deal with the prob_strain NA case at the beginning before any variables
+  ##  are modified
+  if (!is.null(prob_strain) && any(is.na(prob_strain))) {
+    which_nna <- !vapply(seq(ncol(prob_strain)),
```
richfitz, line 69: can use `vlapply` here and drop the `logical(1)`

R/carehomes_rt.R
```diff
@@ -63,6 +63,39 @@ carehomes_Rt <- function(step, S, p, prob_strain = NULL,
...
+                        interpolate_critical_dates, interpolate_min,
+                        eigen_method, R[, which_nna], weight_Rt)
+
+    ## replace reduced step,date,beta with full values (no NA here)
+    ret$step <- step
+    ret$date <- step * p$dt
+    ret$beta <- beta <- sircovid_parameters_beta_expand(step, p$beta_step)
+
+    ## restore full length Rt with NAs when prob_strain is NA
+    for (i in 4:length(ret)) {
```
richfitz, line 84: `for (i in type) {` ?

tests/testthat/test-carehomes-multistrain.R
```diff
@@ -1295,22 +1292,42 @@ test_that("If prob_strain is NA then Rt is NA ", {
   R <- y[index_R, , ]
   prob_strain <- y[index_prob_strain, , ]
 
-  ## all values of prob_strain after the first step should be NA
-  expect_true(all(is.na(prob_strain[, , -1L])))
+  ## set NA for prob_strain in steps 60-70
+  prob_strain[, , 60:70] <- NA
```
richfitz, line 1296: especially given the lintr false positive here, how about:

```r
drop <- 60:70
expect_true(!any(is.na(prob_strain[, , drop])))
expect_true(all(is.na(prob_strain[, , drop])))
```

etc (rename `drop` as you want)

richfitz (changes requested):

R/carehomes_rt.R
RaphaelS1: Doesn't work because `type` may still be `NULL` at this point in the code. I could move it below L109 to resolve this but feel this is easier to read

richfitz: Can we move the `type` resolution above this then? the `4:` make me a bit queasy but it's not a huge deal

annecori: @richfitz  and @RaphaelS1  please check you are happy with my proposed change for this?

richfitz (approved):

annecori (approved):

PR #294, opened 2021-04-23
add carehomes_check_severity

RaphaelS1:
* Adds `carehomes_check_severity` to check severity probs lie in [0,1] after multiplication with rel_ params
* Adds check to end of `carehomes_parameters`
* Exports check for use in ncov


R/support.R
```diff
@@ -611,18 +611,22 @@ combine_rt1_epiestim <- function(what, rt, samples, rank) {
...
+        outer(
+          sircovid_parameters_expand_step(step, p_step[, i]),
+          rel_p[i, , ]
+        )
+      },
+      array(0, c(n_time_steps, n_strains, n_vacc_classes))
+    )
+    assert_proportion(ret)
     ret <- aperm(ret, c(4, 2, 3, 1))
+    assert_proportion(ret)
```
edknock, line 629: we have this appearing twice in the space of three lines

R/support.R
```diff
@@ -637,6 +641,53 @@ compute_pathway_probabilities <- function(step, pars, n_time_steps, n_strains,
...
+##'
+##' @title Check severity parameters
+##'
+##' @param pars A list of probabilities from [carehomes_parameters].
+##'
+##' @return Returns `pars` invisibly if all severity probabilities lie in
+##   (0, 1), otherwise errors.
+##'
+##' @export
+carehomes_check_severity <- function(pars) {
```
edknock, line 654: minor suggestion: since this is carehomes model specific (and features carehomes in the function name), I think this should be moved into carehomes.R (and the corresponding tests moved to test-carehomes-support.R)

R/support.R
RaphaelS1: Neither should have been there!

edknock (approved):

PR #297, opened 2021-04-28
Update to dust 0.9.0

richfitz:
Requires https://github.com/mrc-ide/dust/pull/224

Update branch pin before merge


inst/odin/compare_carehomes.cpp
```diff
@@ -1,4 +1,11 @@
+// Can't use std::min on GPU
+template <typename T>
+T min(const T& a, const T&b) {
```
johnlees, line 3: This needs a HOSTDEVICE I think

johnlees (commented):
The `data_t` in the two models here will be missing the `ALIGN(16)` directive, but I think that's a change in the odin.dust PR

johnlees (approved):

PR #298, opened 2021-04-28
add lag to vaccine schedule

RaphaelS1:
* Adds params `lag_groups` and `lag_days` to control which age groups to add a lag of `lag_days` to for the start of the doses returned by vaccine_schedule

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# sircovid 0.10.49
+
+* Add `lag_groups` and `lag_days` parameters to `vaccine_schedule_future` to manually add lags for give age groups to the vaccine schedule
+
 # sircovid 0.10.47
 
 * New function `sircovid::inflate_state_strains` for adding empty strain compartments to a model state that was run without multiple strains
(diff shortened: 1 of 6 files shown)


NEWS.md
```diff
@@ -1,3 +1,7 @@
+# sircovid 0.10.49
+
+* Add `lag_groups` and `lag_days` parameters to `vaccine_schedule_future` to manually add lags for give age groups to the vaccine schedule
```
annecori, line 3: "given" not "give"

R/vaccination.R
```diff
@@ -409,7 +419,34 @@ vaccine_schedule_future <- function(start,
...
+
+    ## we could do this in apply but loop is more readable
+    for (i in lag_groups) {
+      ## original schedule for group i
+      old_schedule_group_i <- schedule$doses[i, , ]
+      nr <- nrow(old_schedule_group_i)
+      nc <- ncol(old_schedule_group_i)
+
+      ## get start of dose schedule for the group
+      start <- which(old_schedule_group_i[1, ] > 0)[1]
```
annecori, line 437: maybe need a catch for this in case this doesn't exist - e.g. people <50 are not eligible so you won't have any value >0 for them

annecori (changes requested):
looks good, only suggested a change to make sure we catch the case where no vaccine doses are given to an age group

annecori (approved):

PR #300, opened 2021-04-29
Input population

edknock:
Allows a population vector and number of carehome beds to be inputted to `carehomes_parameters`. If these are `NULL` then it sources these for the `region` if we have it within the package.

This still uses UK POLYMOD as the survey in the socialmixr transmission matrix function.


MJomaba (approved):
Looking good.

PR #306, opened 2021-05-07
Booster doses

RaphaelS1:
Boosters can now be added to vaccine schedule.

To-do:

* [x] Update odin
* [x] Add validation check for vaccine_progression_rate (0 for boosters)
* [x] Update vaccine_schedule_future
* [x] Add tests for vacc schedule
* [x] Add function to inflate vacc classes
* [x] Add tests for class inflation

PR #310, opened 2021-05-12
Exclude booster groups

RaphaelS1:
Add parameter to control which age groups do/dont get boosters. Requires #306 to be merged first.


RaphaelS1:
Merged into #306

PR #316, opened 2021-06-09
Compute ALOS

edknock:


RaphaelS1 (commented):
LGTM but looks like tests only check the simple case of no vacc and 1 strain; is it possible to add tests for other cases?

richfitz (approved):
This is fine for code, though I can't comment on the ALOS calculation irtself - @MJomaba do you have more to add

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: sircovid
 Title: SIR Model for COVID-19
-Version: 0.11.9
+Version: 0.11.10
```
richfitz:
```suggestion
Version: 0.11.11
```

richfitz (changes requested):
Ah, after reviewing I noticed the version had moved past us. I've updated the news, and here's a change for the version number

richfitz (approved):

PR #322, opened 2021-06-23
Proper multistrain ifr

edknock:
IFR for a multistrain model was previously only calculating for the 1st strain, and weighting by expected infections by group and vaccine class. This now works for a multistrain model, and we weight by expected infections by group, strain and vaccine class.


R/carehomes_ifr_t.R
```diff
@@ -7,8 +7,8 @@
 ##' @param S A (n groups x n vaccine classes) x steps matrix of
 ##'   "S" compartment counts
 ##'
-##' @param I_weighted A (n groups x n vaccine classes) x steps matrix of
-##'   "I_weighted" compartment counts
+##' @param I_weighted A (n groups x n strains x n vaccine classes) x steps
```
richfitz, line 10: looks like this is still accepted without the strain dimension? Also change matrix -> array

richfitz (commented):
I can't comment on this really. Nothing looks alarming in the code though!

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: sircovid
 Title: SIR Model for COVID-19
-Version: 0.11.14
+Version: 0.11.15
```
annecori, line 3: please can you add to the NEWS file as well? thx

R/carehomes_ifr_t.R
```diff
@@ -76,89 +97,120 @@ carehomes_ifr_t <- function(step, S, I_weighted, p, type = NULL) {
...
+      ## probability of infection by group and vaccine class
+      prob_infection <- (1 - exp(-p$dt * foi))
+      ## expected infections by group and vaccine class
+      expected_infections <- S[, t] * prob_infection
+    } else {
+      ## need to sum over metastrains
+      II <- array(I_weighted[, t], c(n_groups, n_strains, n_vacc_classes))
+      II[, 1, ] <- II[, 1, ] + II[, 4, ]
+      II[, 2, ] <- II[, 2, ] + II[, 3, ]
+      II <- II[, -c(3, 4), , drop = FALSE]
```
annecori, line 134: could/should we add a check that n_strains is indeed 4 and n_real_strain indeed 2 here otherwise may run into issues with all these hard coded numbers?

DESCRIPTION
richfitz: This will need to be .16 now sorry :(

R/carehomes_ifr_t.R
edknock: It's still a matrix, before the 1st dim was of length (n groups x n vaccine class), now it is (n groups x n strains x n vaccine classes)

R/carehomes_ifr_t.R
edknock: Added a check! n_strains = 4 is basically hardcoded into the multistrain version of the model too. But good idea, add this in so we know this needs changing if the multistrain part changes again

annecori (approved):

PR #325, opened 2021-06-24
allow second strain rel param < 1

RaphaelS1:


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: sircovid
 Title: SIR Model for COVID-19
-Version: 0.11.15
+Version: 0.11.16
```
annecori, line 3: depending when this is merged may need to upgrade again because of Ed's PR on IFR

annecori (approved):

PR #332, opened 2021-07-08
add p_R

edknock:


tests/testthat/test-carehomes-check.R
```diff
@@ -905,22 +978,3 @@ test_that("Individuals can infect in compartment with non-zero transmission", {
   helper("ICU_transmission", "ICU_W_R_unconf", "gamma_ICU_W_R_step")
   helper("ICU_transmission", "ICU_W_R_conf", "gamma_ICU_W_R_step")
 })
-
-
-test_that("No one is hospitalised, no-one recovers in edge case", {
```
edknock, line 910: Have removed this test because it seems to be testing the same thing as the "No one dies in the community if p_G_D is 0" test

RaphaelS1 (approved):
Missing updated NEWS otherwise looks good to me

richfitz (approved):

PR #338, opened 2021-07-21
introduce p_NC_weekend

edknock:


R/carehomes.R
```diff
@@ -375,6 +378,7 @@ carehomes_parameters <- function(start_date, region,
                                  m_CHW = 4e-6,
                                  m_CHR = 5e-5,
                                  p_NC = 0.01,
+                                 p_NC_weekend = 0.005,
```
richfitz, line 381: does this make more sense parametrised relative to p_NC (i.e. a weekend effect of 0.5 for this combination)

R/carehomes.R
```diff
@@ -614,14 +619,15 @@ carehomes_index <- function(info) {
...
   ## Only incidence versions for the likelihood now:
-  index_run <- index_core[c("icu", "general", "deaths_carehomes_inc",
-                            "deaths_comm_inc", "deaths_hosp_inc",
-                            "admitted_inc", "diagnoses_inc",
-                            "sero_pos_1", "sero_pos_2", "sympt_cases_inc",
-                            "sympt_cases_non_variant_inc",
-                            "sympt_cases_over25_inc",
-                            "sympt_cases_non_variant_over25_inc",
-                            "react_pos")]
+  index_run <- c(time = index[["time"]],
```
richfitz, line 622: I think this should be added to index_core, unless there's something I have forgotten

R/carehomes.R
```diff
@@ -726,16 +732,19 @@ carehomes_compare <- function(state, observed, pars) {
 
   ## calculate test positive probabilities for the various test data streams
   ## Pillar 2
-  pillar2_negs <- pars$p_NC * (pars$N_tot_all - model_sympt_cases)
+  time <- state["time", ]
+  p_NC <- ifelse(grepl("^S", weekdays(sircovid_date_as_date(time))),
```
richfitz, line 736: I'd recommend using the weird modulo arithmetic approach I posted before, as we'll need that for the C++ version

richfitz (changes requested):

R/carehomes.R
edknock: I put `time` in `index_run` but not `index_core` so it can be used in the compare but not get saved

richfitz: ah, cool - that's what I've forgotten. Perhaps make a note? The now-old comment above index_core mislead me a bit

R/carehomes.R
edknock: I guess if you wanted to force `p_NC_weekend` to be less than `p_NC`, but perhaps I'd rather let the data determine that. Also I think it might be better from a fitting POV to fit `p_NC` and `p_NC_weekend`, as these would be less correlated than `p_NC` and some multiplicative weekend effect

R/carehomes.R
edknock: Have switched to the modulo arithmetic approach, agree it's better if consistent between the two

richfitz (approved):

PR #351, opened 2021-09-30
Lancelot remove hosp testing

edknock:


edknock:
Master has long since diverged from this. Maybe will revisit later on

PR #357, opened 2021-11-03
Update to dust 0.11.0

richfitz:
@pabloperguz , this is why tests are starting to fail, we merged a big dust PR recently


DESCRIPTION
```diff
@@ -55,19 +55,19 @@ LazyData: true
 Language: en-GB
 Depends: R (>= 3.1.0)
 Imports:
-    dust (>= 0.10.0),
+    dust (>= 0.11.0),
     eigen1 (>= 0.1.1),
     mcstate (>= 0.6.8),
```
edknock, line 60: Do you want to bump this to 0.6.10?

richfitz: good idea, I've set that now

edknock (approved):

PR #361, opened 2021-11-12
Refactor vaccine schedule from data

edknock:
This refactors `vaccine_schedule_from_data` and removes `vaccine_schedule_data_future`. `vaccine_schedule_from_data` now takes data that can have any number of dose columns, and can include rows where `age_band_min` is `NA` - those will be considered doses to be distributed across all ages according to the priority schedule (after all age-specific doses are taken care of).

The `uptake` input to this now has a second dimension representing doses - so that uptake can decrease as dose number increases.

Future vaccine scheduling will be refactored in a subsequent PR. `vaccine_schedule_data_future` is removed so that we just use data in the fits, and then use "future" scheduling in any simulations. The averaging of the last 7 days of doses has thus been removed - this is because we now shift doses by at least a week, so any doses in the last 7 days will actually be implemented in simulations. So I think therefore it is better to fill those in by just using `doses_future` in simulations.

Due to removal of this function, I have removed a test which results in two extra codecov misses. This test is for when `doses_future` in `vaccine_schedule_scenario` is NULL - but I do not think we have ever used this functionality, so this can probably be removed when future schedule functions are refactored.

UPDATE: codecov of the above has been dealt with by instead throwing an error in that instance (and testing for that error).


R/vaccination.R
```diff
@@ -506,108 +506,140 @@ vaccine_schedule <- function(date, doses, n_doses = 2L) {
...
+  age_start <- sircovid_age_bins()$start
+  if (nrow(uptake) != length(age_start) + 2) {
+    stop(sprintf("Expected uptake to have %s rows as there are %s groups",
+                 length(age_start) + 2, length(age_start) + 2))
+  }
+  if (ncol(uptake) != n_doses) {
+    stop(sprintf("Data has %s dose columns so expected uptake to have %s
+                 columns", n_doses, n_doses))
+  }
+
```
annecori, line 558: should we also check that uptake is decreasing as dose index increases?

R/vaccination.R
```diff
@@ -506,108 +506,140 @@ vaccine_schedule <- function(date, doses, n_doses = 2L) {
...
+    f <- function(dose) {
+      vaccination_schedule_exec(daily_doses_tt, agg_doses[dose, ],
+                                population_left,
+                                population_to_vaccinate_mat,
+                                Inf,
+                                0L, dose)
+    }
+
+    doses <- doses +
```
annecori, line 637: I never know what this Reduce thing does and haven't checked the indexes of the apply but assuming this is ok

annecori (approved):
Code looks good to me, I only have a couple of minor comments/suggestions

R/vaccination.R
edknock: reduces a list to a single value (by summing the items in the list in this case)

R/vaccination.R
edknock: good idea, have added that, plus a check that uptake is all in [0,1]

annecori (approved):
looks good to me with new checks!

richfitz (approved):

PR #364, opened 2021-11-17
Boost before wane

edknock:
This adds support into the `lancelot` model for a "vaccine skip move" - with the purpose of allowing people to get a booster before they have waned.

The vaccine skip move is from vaccine stratum `vacc_skip_from` to `vacc_skip_to`. We require that `vacc_skip_from <= vacc_skip_to` (this is checked in `lancelot_parameters`). If there is only one vaccine stratum then both of these parameters must be 1. There is a base progression rate for the vaccine skip move, which is used unless the standard vaccine move from `vacc_skip_to - 1` to `vacc_skip_to` is controlled by a dose - `vacc_skip_dose` takes the corresponding number of that dose (otherwise it is 0).

Doses are distributed within the odin code, and we have a parameter `vacc_skip_weight` which must be in [0, 1] - this represents how individuals in `vacc_skip_from` are weighted for dose distribution relative to those who are just making a "vaccine next class" move. If there are excess doses for the "vaccine next class", then those are redistributed to the vaccine skip move candidates. Missed doses are carried forward as before. Note that it `vacc_skip_weight = 0`, then no doses are available for vaccine skip moves at all (allowing us to switch off the vaccine skip move).

NOTE: buildkite is failing due to an issue that @richfitz will fix.


R/lancelot.R
```diff
@@ -390,6 +410,10 @@ lancelot_parameters <- function(start_date, region,
                                 vaccine_index_booster = NULL,
                                 vaccine_catchup_fraction = 1,
                                 n_doses = 2L,
```
annecori, line 412: slightly unrelated to the PR but should we make n_doses = 3 the default now?

R/lancelot.R
```diff
@@ -497,6 +521,40 @@ lancelot_parameters <- function(start_date, region,
...
+  }
+  if (vaccination$n_vacc_classes == 1) {
+    if (vacc_skip_from != 1) {
+      stop("n_vacc_classes = 1 so require vacc_skip_from = 1")
+    }
+    if (vacc_skip_to != 1) {
+      stop("n_vacc_classes = 1 so require vacc_skip_to = 1")
+    }
+  } else {
+    if (vacc_skip_to < vacc_skip_from) {
```
annecori, line 544: should we add something else which checks that 
if (vacc_skip_to == vacc_skip_from) vacc_skip_progression_rate should be zero?

annecori: (otherwise we do a lot of work for nothing since we move people into their own compartment...

inst/odin/lancelot.R
```diff
@@ -20,16 +20,27 @@ update(time) <- (step + 1) * dt
 ## output number of individuals vaccinated by age and vaccine stage
 ## For example, for E, we sum over n_E_next_vacc_class (those moving vaccine
 ## stage without progressing disease stages)
-## vaccinated S
-n_S_vaccinated[, ] <- n_S_next_vacc_class[i, j]
+## Note that individuals making vaccine skip moves are counted in all classes
```
annecori, line 23: not sure what this comment means @edknock ?

annecori: ah now I remember this discussion - worth expanding a bit I think to explain this is only for the output it doesn't affect the model dynamics

inst/odin/lancelot.R
```diff
@@ -20,16 +20,27 @@ update(time) <- (step + 1) * dt
...
+## they skip over
+n_S_vaccinated[, ] <- n_S_next_vacc_class[i, j] +
+  (if (j >= vacc_skip_from && j < vacc_skip_to) n_S_vacc_skip[i]
+   else 0)
```
annecori, line 27: the code is ok but maybe the comment could be a bit clearer. e.g. for those who skip a class, in the outputs, we count them as if they had gone through all the successive classes (possibly without staying any time in the intermediate classes) - I think that would be a bit clearer

inst/odin/lancelot.R
```diff
@@ -66,6 +77,38 @@ initial(cum_n_vaccinated[, ]) <- 0
 update(cum_n_vaccinated[, ]) <- cum_n_vaccinated[i, j] + n_vaccinated[i, j]
 dim(cum_n_vaccinated) <- c(n_groups, n_vacc_classes)
 
+
```
annecori, line 80: again maybe adding more comments in code to say because we counted the people who skip classes together with the others in the outputted numbers of  people who have progressed to a given vaccine class, but we may want to monitor the skipping vs non skipping flow separately, we are also outputting the skipping only

inst/odin/lancelot.R
```diff
@@ -1746,16 +1850,36 @@ vaccine_probability_doses[, ] <- min(
   as.numeric(1))
 dim(vaccine_probability_doses) <- c(n_groups, n_doses)
 
-vaccine_attempted_doses[, ] <- vaccine_missed_doses[i, j] + (
-  if (as.integer(step) >= dim(vaccine_dose_step, 3) ||
-      vaccine_n_candidates[i, j] == 0) 0
+tmp_attempted_doses[, ] <- vaccine_missed_doses[i, j] + (
```
annecori, line 1853: I'd suggest renaming tmp_attempted_doses to total_attempted_doses
then clarify that you split the total into the "direct" and the "skip" parts

inst/odin/lancelot.R
```diff
@@ -1746,16 +1850,36 @@ vaccine_probability_doses[, ] <- min(
...
+  (if (j == vacc_skip_dose)
+    min(vaccine_n_candidates[i, j] /
+          (vaccine_n_candidates[i, j] +
+             vacc_skip_weight * vacc_skip_n_candidates[i])
+        * tmp_attempted_doses[i, j],
+        vaccine_n_candidates[i, j])
+    else tmp_attempted_doses[i, j])
 dim(vaccine_attempted_doses) <- c(n_groups, n_doses)
 
+vacc_skip_attempted_doses[] <-
```
annecori, line 1870: it would be nice to use an extreme case to check if we ever run out of people to vaccinate in the "skip" route whilst not having vaccinated all candidates in the "direct" route.

tests/testthat/test-lancelot-vaccination.R
```diff
@@ -3220,3 +3220,222 @@ test_that("run sensible vaccination schedule with waning and boosters", {
...
+  expect_error(lancelot_parameters(1, "london", rel_susceptibility = c(1, 1, 1),
+                                   vacc_skip_from = 1L, vacc_skip_to = 3L,
+                                   vacc_skip_weight = 1.4),
+               "'vacc_skip_weight' must lie in")
+
+  expect_error(lancelot_parameters(1, "london", rel_susceptibility = c(1, 1, 1),
+                                   vacc_skip_from = 1L, vacc_skip_to = 3L,
+                                   vacc_skip_weight = -1),
+               "'vacc_skip_weight' must lie in")
+  expect_error(lancelot_parameters(1, "london", rel_susceptibility = c(1, 1, 1),
```
annecori, line 3249: I think you've got the exact same test repeated twice?

tests/testthat/test-lancelot-vaccination.R
```diff
@@ -3220,3 +3220,222 @@ test_that("run sensible vaccination schedule with waning and boosters", {
...
+
+  state <- lancelot_initial(info, 1, p)$state
+
+  mod$update_state(state = state)
+
+  set.seed(1)
+  y <- mod$simulate((1:200) * 4)
+  y <- mod$transform_variables(drop(y))
+
+  ## All booster moves are via vaccine skip
```
annecori, line 3373: could also check here that this is not zero i.e. people are indeed receiving boosters

annecori (changes requested):
Looks great; I've only made a few suggestions mainly about commenting more and adding one small bit in a test

R/lancelot.R
edknock: Possibly, but not essential. Would likely require a lot of changes to tests so could be done in its own PR!

tests/testthat/test-lancelot-vaccination.R
edknock: Removed, thanks

inst/odin/lancelot.R
edknock: I'm not sure how to test it but it should be mathematically impossible given `vacc_skip_weight <= 1`

edknock: Let's say there are `x` unwaned candidates, `y` waned candidates, `z` doses to share out between them and the unwaned are weighted `w <= 1` relative to waned.

So we give `w * x * z / (w * x + y)` doses to the unwaned and `y * z/ (w * x + y)` doses to the waned. 

To have an excess of doses for the unwaned while not all waned individuals are vaccinated we need
`w * x * z / (w * x + y) > x` (1)
and
`y * z/ (w * x + y) > y` (2)

Rearranging (2) gives us `z < w * x + y` (3)

Rearranging (1) gives us `z > (w * x + y) / w`. Given w <= 1, then `z > w * x + y` - and since this contradicts (3) the situation is impossible

inst/odin/lancelot.R
edknock: I've changed this, and added comments

tests/testthat/test-lancelot-vaccination.R
edknock: Done

inst/odin/lancelot.R
edknock: I've expanded this

inst/odin/lancelot.R
edknock: Hopefully a bit clearer now

inst/odin/lancelot.R
edknock: Added a comment about this

inst/odin/lancelot.R
annecori: ok! I think you wrote (2) with ">" instead of "<" but the rest of the maths is ok so happy with that thanks @edknock

R/lancelot.R
annecori: @edknock just bringing this up as I'm not sure it's been addressed? Also one thing I wondered about was whether we should check that vacc_skip_to >= vacc_skip_from + 2 since we want to have a check (except when n_doses = 1 maybe where we allow moves within the same compartment?). let mw know what you think

richfitz:
On the buildkite failure: I know where this is coming from and it affects only GPU code generation. I'm having a look at a fix but it will be in odin.dust with no change needed here

R/lancelot.R
edknock: I have revamped the checking so it is better overall, dealing with these issues

annecori (approved):

PR #366, opened 2021-11-23
i222: initial seeding

edknock:
This changes seeding so that instead of starting on `start_date` with `initial_I` asymptomatic 15-19 year olds, we start from step 0 and seed from S compartment to E compartment from the `start_date`. We have an `initial_seed_size`, which can then be split over steps using `initial_seed_pattern`. `start_date` can now be non-integer, and if it falls between steps then the seeding is adjusted and we split it over an additional step.

The strain is also now seeded in a similar way, with `strain_seed_date`, `strain_seed_size` and `strain_seed_pattern`.

We now have an issue where we run the model for a bit at the beginning without any infectives (both before the seeding and after the seeding but before any seeded individuals have become infectious) - previously this would've given us NAs for weightings used for Rt and IFR calculations, hence outputting NAs. I have decided to avoid the NA weighting, and thus if there are no infectives (sum(I_weighted) = 0) then we default to `prob_strain = 1` (so we weight 100% the Rt for strain 1) and `I_weighted` adds some weight only in the 15-19 yo, unvaccinated, strain 1 group.


R/basic.R
```diff
@@ -207,28 +212,13 @@ basic_initial <- function(info, n_particles, pars) {
...
   index_N_tot <- index[["N_tot"]]
 
-  state[index_S] <- initial_S
-  state[index_I] <- initial_I
+  state[index_S] <- pars$population
   state[index_N_tot] <- sum(pars$population)
 
-  list(state = state,
-       step = pars$initial_step)
+  state
```
edknock, line 221: @richfitz do we want to change this back to a list of `state` and `step` and we hardcode the initial step here?

richfitz: yes please

R/lancelot.R
```diff
@@ -1553,37 +1532,31 @@ lancelot_parameters_strain <- function(strain_transmission, strain_seed_date,
...
-        i <- seq.int(
-          strain_seed_date[[j]] / dt,
-          strain_seed_date[[j + 1]] / dt - 1
-        )
-      }
-      strain_seed_step[i] <- strain_seed_rate[[j]] * dt
-    }
+      c(strain_seed_step <- numeric(floor(initial_strain_seed_step)),
+        strain_seed_size *
+          seed_over_steps(initial_strain_seed_step, strain_seed_pattern), 0)
```
richfitz, line 1559: This zero should not be needed? Or is this not updated to use the new offset approach? Could update that later perhaps

R/parameters.R
```diff
@@ -17,18 +18,30 @@ sircovid_parameters_shared <- function(start_date, region,
...
+  if (is.null(initial_seed_pattern)) {
+    initial_seed_pattern <- 1.0
+  }
+
+  start_step <- start_date * steps_per_day
+  initial_seed_step <- floor(start_step)
+  initial_seed_value <-
+    seed_over_steps(start_step, initial_seed_pattern) * initial_seed_size
+
+  ## TODO: initial_step can come out here once lancelot updated, no
```
richfitz, line 32: is this now resolved?

inst/odin/lancelot.R
```diff
@@ -366,6 +366,21 @@ n_S_progress[, , ] <- if (j == 1 || n_real_strains == 1)
   rbinom(n_S_progress_tot[i, k], rel_foi_strain[i, j, k]) else
     (if (j == 2) n_S_progress_tot[i, k] - n_S_progress[i, 1, k] else 0)
 
+## Seeding of first wave, we seed in group 4, strain 1, vaccine stratum 1
+##
+seed_step_end <- seed_step_start + length(seed_value)
+seed <- if (step >= seed_step_start && step < seed_step_end)
```
richfitz, line 372: consider 
```
seed_mean <- if...
seed <- rpois(seed_mean)
```

then simplifying call to `n_S_progress`

richfitz (commented):
Really surprised how little this affected the tests

R/lancelot.R
edknock: Updated the odin approach for seeding the strain so this is outdated

inst/odin/lancelot.R
edknock: have gone with `seed_rate` (we already have `strain_seed_rate`)

inst/odin/lancelot.R
```diff
@@ -366,6 +366,22 @@ n_S_progress[, , ] <- if (j == 1 || n_real_strains == 1)
   rbinom(n_S_progress_tot[i, k], rel_foi_strain[i, j, k]) else
     (if (j == 2) n_S_progress_tot[i, k] - n_S_progress[i, 1, k] else 0)
 
+## Seeding of first wave, we seed in group 4, strain 1, vaccine stratum 1
```
richfitz, line 369: This is great, really tidy

richfitz (approved):
This is ready for someone else to look at from the science perspective. I expect a bit of a fiddle will be required to chase this through spimalot and the fits due to the change in initial conditions, but it might not be all that bad

R/parameters.R
edknock: I decided to use the initial_step here and left it in - have removed the TODO comment

MJomaba (approved):
Code looking good to me.

PR #367, opened 2021-11-24
Update lancelot_rt.R

epiyasin:
ensuring that interpolate_critical_index can be safely parsed without NAs and with correct order


pabloperguz (approved):
Code looks good and have tested the whole pipeline my end. Feel free to merge once all tests have passed

PR #369, opened 2021-11-30
Patch rt cross immunity

edknock:
Patches up the Rt calculation wrt to cross immunity being implemented the wrong way round. Would only have affected the effective Rt calculated in post-processing (would not affect the Rt excluding immunity)


inst/odin/lancelot.R
```diff
@@ -477,6 +477,10 @@ n_I_P_vacc_skip[, , ] <-
 ##  R1 and R2 can progress to S (w.r. waning_rate[i]) or
 ##  R1 can progress to E3 (w.r. strain 2 (3 - 1))
 ##  R2 can progress tp E4 (w.r. strain 1 (3 - 2))
+##
+## Note that (if n_strains > 2)
```
edknock, line 481:
```suggestion
## Note that (if n_real_strains = 2)
```

annecori (approved):

PR #373, opened 2021-12-09
Add support for rotating strains

richfitz:
Taken from @annecori's code in ncov-outputs, adds tests in a similar style to our inflation functions.

The name can be changed if changed in **this** PR otherwise it will be annoying. It's not really rotation because we are clearing out strain 2 but no better word jumps out at me.

Done while pairing with @Divyakanapram, so all questions to her :grinning:


R/tools.R
```diff
@@ -182,3 +182,109 @@ check_sircovid_model <- function(sircovid_model) {
...
+  strain_to_idx <- 1
+
+  ## TODO: someone please check this with the odin code. How would we
+  ## know if additional compartments are added that should be put
+  ## here?  In contrast with inflate_state which works by detecting
+  ## which states have changed size, we can't easily do that here.
+  ##
+  ## It's possible that something could be derived from the odin code
+  ## by looking for the expression that underpins each *variable* and
+  ## seeing if it contains n_strains. Rich can do this.
```
annecori, line 229: I wondered about this and I don't know how to solve this, at the moment I did all this manually I'm afraid. Leaving this with you @richfitz I think...

annecori (approved):

PR #385, opened 2022-01-07
Tidy up data processing

richfitz:
We're awkwardly split across spimalot and sircovid (and actual use) here, so time to have a tidy before adding multiregion into the mix

* delete a couple of exported functions that are no longer used outside tests `lancelot_particle_filter`, `sircovid_data`
* remove some internal functions `lancelot_data`, `basic_data`
* rename `lancelot_prepare_data` to `lancelot_check_data` as it does not modify the data
* rewrite the internals of `lancelot_check_data` to be more readable, and reduce duplication when using multiregion data

Will be used by https://github.com/mrc-ide/spimalot/pull/161


richfitz:
The codecov failure here is just that I deleted a bunch of code so the project percentage goes down

edknock (approved):

PR #386, opened 2022-01-10
Remove IFR post-processing calculation

edknock:
Removes the code for calculating IFR in post-processing, removing the need for `I_weighted` being outputted. All code used only for the IFR calculation has been removed.

IFR will be added to the odin code in a subsequent PR.


edknock:
Codecov is failing due to the existing miss on master, which means that because code has been removed the % coverage has fallen despite no new misses have being introduced.

epiyasin (approved):

PR #390, opened 2022-01-13
Drop unused functions

richfitz:
These are no longer used in spimalot or ncov-outputs

Amazingly, this removes that annoying 1-line coverage hole that has been haunting us for ages, that nobody wanted to fix


edknock (approved):
Good to see code coverage back at 100%!

PR #393, opened 2022-01-18
gamma_U is now time-varying

edknock:
This allows the time to testing positive in hospital to vary over time, so `gamma_U` becomes `gamma_U_step`


epiyasin (approved):
Code looks fine. Tested with no issues, merging.

PR #419, opened 2022-06-23
Store historic variants in R5

edknock:
This introduces a 5th layer in to the R compartment when we are running the two strains model. R5 now means "recovered from historic variants". There is no movement into this layer within the odin code, however when we rotate strains:

- anyone in R1 or R4 will now move into R5
- R2 and R3 will move into R1 and R4 respectively
- In other compartments with a strain dimension, those in X1 and X4 will remain where they are and X2 and X3 will move into X1 and X4 respectively

Then within the odin code movement from R to E is now as follows:
- Anyone in R5 can move to R3 (strain 2 reinfection) or R4 (strain 1 reinfection)
- Anyone in R1/R4 (recovered from strain 1 infection/reinfection) can move to R3 (strain 2 reinfection)
- Anyone in R2/R3 (recovered from strain 2 infection/reinfection) is entirely immune to strain 1

This essentially means that people who have recovered from a variant cannot be reinfected by another variant that predated it. This is likely fine as generally strain 2 becomes dominant fairly quickly.

Note that `cross_immunity` retains its inverted indexing - now it essentially means c(cross immunity against strain 2 given by strains that predate strain 2, cross immunity against strain 1 given by strains that predate strain 1).

Note that the vaccine delay paper model should be a submodel of this model.


R/tools.R
```diff
@@ -222,8 +222,8 @@ rotate_strains <- function(state, info, ...) {
   ## infected in the past is moved to now be indexed by strain 1 (no
```
pabloperguz, line 222: This comment would need updating

pabloperguz (commented):
Nice work! Just a very minor thing wrt an outdated comment

pabloperguz (approved):

PR #421, opened 2022-06-27
Reinfections go to T_PCR_pre instead of T_PCR_pos

edknock:


inst/odin/lancelot.R
```diff
@@ -1063,13 +1063,13 @@ new_R[, , ] <- R[i, j, k] -
 ## Work out the PCR positivity
 new_T_PCR_pre[, , , ] <- T_PCR_pre[i, j, k, l] -
   n_T_PCR_pre_progress[i, j, k, l] +
-  (if (k == 1) n_S_progress[i, j, l] else
+  (if (k == 1) n_S_progress[i, j, l] +
+     (if (j > 2) n_RE[i, j - 2, l] else 0) else
```
edknock, line 1067: Note: `n_RE[i, j, l]` is actually the number going from strain j to strain j+2. No-one going from R to E goes to strain 1 or 2, strain 1 will go to strain 3 and strain 2 will go to strain 4.

pabloperguz (approved):

PR #435, opened 2022-12-09
Simpler check severity

edknock:


R/lancelot.R
```diff
@@ -2774,29 +2774,18 @@ lancelot_n_groups <- function() {
...
-        if (is.null(pars[[rel_p]])) {
-          stop(sprintf("Parameter '%s' is missing", rel_p))
-        } else if (is.null(pars[[p_step]])) {
-          stop(sprintf("Parameter '%s' is missing", p_step))
-        }
+    if (is.null(pars[[rel_p]])) {
+      stop(sprintf("Parameter '%s' is missing", rel_p))
+    } else if (is.null(pars[[p_step]])) {
+      stop(sprintf("Parameter '%s' is missing", p_step))
+    }
```
edknock, line 2781: These tests were being applied over $i$ even though they do not depend upon $i$. Thus they were being applied $n_groups$ times. Here we are now applying the same tests just once

R/lancelot.R
```diff
@@ -2774,29 +2774,18 @@ lancelot_n_groups <- function() {
...
 
-        ## simple but crude checks
-        min_p <- min(pars[[rel_p]][i, , ]) * min(pars[[p_step]][, i])
+    if (!(ncol(pars[[rel_p]]) %in% c(1, 4))) {
+      stop(sprintf("%s should have 1 or 4 columns", rel_p))
+    }
```
edknock, line 2785: Again, this test only needs to be applied once

R/lancelot.R
```diff
@@ -2774,29 +2774,18 @@ lancelot_n_groups <- function() {
...
-        } else {
-          if (!(ncol(pars[[rel_p]]) %in% c(1, 4))) {
-            stop(sprintf("%s should have 1 or 4 columns", rel_p))
-          }
-        }
-        TRUE
-      }, logical(1)
-    )
+    assert_non_negative(pars[[rel_p]], rel_p)
+    assert_non_negative(pars[[p_step]], p_step)
```
edknock, line 2788: These checks will ensure we error whenever we would have errored before, but are single checks across a whole array rather than n_groups slices. Essentially the same behaviour as before, but with different error messages

R/parameters.R
```diff
@@ -129,8 +129,9 @@ sircovid_parameters_piecewise_linear <- function(date, value, dt) {
     value <- rbind(value[1, ], value)
   }
 
+  times <- seq(0, date[[length(date)]], by = dt)
   value <- apply(value, 2, function(x) {
-    stats::approx(date, x, seq(0, date[[length(date)]], by = dt))$y
+    stats::approx(date, x, times)$y
```
edknock, line 134: this works the same as before, it just takes the `seq` out of the apply - it does not depend upon

R/parameters.R
```diff
@@ -216,8 +217,9 @@ sircovid_parameters_piecewise_constant <- function(date, value, dt) {
     }
   }
 
+  times <- seq(0, date[[length(date)]], by = dt)
   stats::approx(date, value, method = "constant",
-                xout = seq(0, date[[length(date)]], by = dt))$y
+                xout = times)$y
```
edknock, line 222: Same as above

tests/testthat/test-lancelot-support.R
```diff
@@ -1487,10 +1487,18 @@ test_that("lancelot check severity works as expected", {
...
     )
     p[[rels[[i]]]][] <- 1
+
+    p[[steps[[i]]]][] <- -1
+    expect_error(
+      lancelot_check_severity(p),
+      sprintf("'%s' must have only non-negative values", steps[[i]]),
+      fixed = TRUE
+    )
+    p[[steps[[i]]]][] <- 1
```
edknock, line 1501: Just an additional test - note all other tests passing unchanged!

tests/testthat/test-lancelot-support.R
```diff
@@ -1487,10 +1487,18 @@ test_that("lancelot check severity works as expected", {
     p[[rels[[i]]]][] <- -1
     expect_error(
       lancelot_check_severity(p),
-      sprintf("%s * %s is not in [0, 1] for group 1", rels[[i]], steps[[i]]),
+      sprintf("'%s' must have only non-negative values", rels[[i]]),
```
edknock, line 1490: This is just because the error message has changed - but the same case still produces an error!

MJomaba (approved):
Code makes sense and looks good.

